Issue #302, opened 2014-08-19
one mapping for two fields


ivan1986:
I add two fields in one entity whis one mapping dir

```
    /** @Vich\UploadableField(mapping="questions", fileNameProperty="soundName") */
    protected $sound;

    /** @Vich\UploadableField(mapping="questions", fileNameProperty="backName") */
    protected $back;
```

second not work, this may be bug or know issue

PR #355, opened 2014-12-09
Allow the same mapping to be used more than once

K-Phoen:
Fixes issue #302 and closes PR #305.

From now, internal services work with `PropertyMapping` objects and the userland API asks for field names.
The behavior change introduced in v0.10.0 was broken and this behavior should be considered as definitive once this PR is merged.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,7 +1,10 @@
 Upgrading from v0.12.0 to master
 ================================
 
-Nothing!
+- the `resolvePath` and `resolveUri` Storage methods now take a field name
+  instead of a mapping name. The same goes for the UploaderExtension and
+  UploaderHelper `asset` method.
+  This reverts the broken behavior introduced in v0.10.0.
 
 Upgrading from v0.11.0 to v0.12.0
 ================================


webdevilopers:
Sorry for commenting on a merged PR but I still seam to have this problem with current `dev-master` with the following annotations:
```php
    /**
     * @Vich\UploadableField(mapping="ventura_configuration_file", fileNameProperty="file1Name")
     *
     * @var File $file1
     */
    private $file1;

    /**
     * @var file1Name
     *
     * @ORM\Column(type="string", length=255, name="file1_name", nullable=true)
     */
    private $file1Name;

    /**
     * @Vich\UploadableField(mapping="ventura_configuration_file", fileNameProperty="file2Name")
     *
     * @var File $file2
     */
    private $file2;

    /**
     * @var file2Name
     *
     * @ORM\Column(type="string", length=255, name="file2_name", nullable=true)
     */
    private $file2Name;

    public function getFile1()
    {
        return $this->file1;
    }

    public function setFile1(File $file1 = null)
    {
        $this->file1 = $file1;
    }

    public function getFile1Name()
    {
        return $this->file1Name;
    }

    public function setFile1Name($file1Name)
    {
        $this->file1Name = $file1Name;
    }

    public function getFile2()
    {
        return $this->file2;
    }

    public function setFile2(File $file2 = null)
    {
        $this->file2 = $file2;
    }

    public function getFile2Name()
    {
        return $this->file2Name;
    }

    public function setFile2Name($file2Name)
    {
        $this->file2Name = $file2Name;
    }
```

After submitting the form my entity is populated in a strange way:
```
  -file1: File {#2519}
  -file1Name: "5572032d5f68b.JPG"
  -file2: UploadedFile {#14
    -test: false
    -originalName: "Phoenix-1024x422.jpg"
    -mimeType: "image/jpeg"
    -size: 64287
    -error: 0
  }
  -file2Name: null
```

Only `file1` is stored in database and file system.

PR #386, opened 2015-02-25
Fix error with annotation using Doctrine ORM

bocharsky-bw:
The `@note` annotation thrown exception `Doctrine\Common\Annotations\AnnotationException`:

```
[Semantical Error] The annotation "@note" in property Acme\DemoBundle\Entity\Image::$imageFile was never imported. Did you maybe forget to add a "use" statement for this annotation?
```


K-Phoen:
Thanks :)

PR #387, opened 2015-02-25
Replace form.resources with form_themes key

bocharsky-bw:
The `twig.form.resources` Twig config key is deprecated.

### DIFF
--- a/Resources/doc/form/vich_file_type.md
+++ b/Resources/doc/form/vich_file_type.md
@@ -26,16 +26,16 @@ with VichUploaderBundle.
 ```yaml
 # app/config/config.yml
 twig:
-    # ...
-    form:
-        resources:
-            # other form themes
-            - 'VichUploaderBundle:Form:fields.html.twig'
+    form_themes:
+        # other form themes
+        - 'VichUploaderBundle:Form:fields.html.twig'
 ```
+
+> **N.B:** the `form_themes` parameter is introduced in Symfony 2.5, check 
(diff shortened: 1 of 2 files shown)


K-Phoen:
We might want to document both `form.resources` and `form_themes` as this bundle is compatible with Symfony ~2.3 and `form_themes` isn't available before 2.5.

Or we could just skip this section and give a link to the documentation.

What do you think?

bocharsky-bw:
@K-Phoen I think better to show latest available config example in docs.
Also we have link to the official docs, so if someone using an older version of Symfony - he can go to the official docs and see config example for his version. Official Symfony docs have a great versioning until Symfony `2.0`.

So I vote for a new config example with `form_themes` key.

K-Phoen:
Hum. In this case, I'd go for the up-to-date example (`form_themes`) with a note telling users that if they use Symfony < 2.5 they should read [the doc](http://symfony.com/doc/2.3/cookbook/form/form_customization.html#php)

Could look like this:

> **N.B:** the `form_themes` parameter is introduced in Symfony 2.5, check [the documentation](http://symfony.com/doc/2.3/cookbook/form/form_customization.html#php) if you use an older version.

bocharsky-bw:
Yes, you're right.
It would be better with this note
Thanks!

K-Phoen:
Can you add this note in your PR and squash your commits?

I'll merge it once it's done :)

bocharsky-bw:
Ok, I've done

K-Phoen:
Perfect, thanks!

PR #405, opened 2015-06-26
Fix default filename suffix usage

Restless-ET:


K-Phoen:
Thanks for your contribution!

A unit test for this could be nice. Can you write it?
I'll write one before merging if you can't.

Restless-ET:
Added a unit test for a custom filename suffix.

K-Phoen:
:heart:

Restless-ET:
:smile: 

BTW, for when are you planning a new tag release?

K-Phoen:
There are two PR and a few issues to review before that but it should be soon (less than a week).

PR #411, opened 2015-06-30
Created helpful bash script for running tests locally.

Restless-ET:
I've created this small bash script when I was developing the unit test for #405 to help me run the tests against all the Symfony versions supported by the bundle.

I've made a few adjustments today and decided to add it to my fork.
I'm creating this PR in case you're interested in adding it to the bundle.

It supports passing a single SF2 version and/or the arguments for phpunit. Example:
```
~$ SYMFONY_VERSION=2.7.1 ./runTests.sh --filter testCustomFileNameProperty
```

PS- I've only tested this in Ubuntu 14.04 with the packaged PHP version (5.5.9).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,7 +40,7 @@ PHPUnit should be installed by composer. Run the tests with the
 Alternatively you can use the `runTests.sh` bash script present at the project root.
 Default usage example:
 
-```~$ ./runTest.sh``` (This runs all tests with your current PHP version against all the support Symfony versions.)
+```~$ ./runTest.sh``` (This runs all tests with your current PHP version against all supported Symfony versions.)
 
 You can also set a specific Symfony version to test against and/or pass the arguments for PHPUnit
  as arguments to the script. Usage example:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,7 +32,19 @@ Thank you!
 ## Running the test suite
 
 Ensure that the required vendors are installed by running `composer install`.
-The test suite requires the `php5-mongo` extension to be installed.
+The test suite requires the `php5-mongo` and `php5-sqlite` extensions to be installed.
 
 PHPUnit should be installed by composer. Run the tests with the
 `./vendor/bin/phpunit` command.
+
+Alternatively you can use the `runTests.sh` bash script present at the project root.
+Default usage example:
+
+```~$ ./runTest.sh``` (This runs all tests with your current PHP version against all the support Symfony versions.)
+
+You can also set a specific Symfony version to test against and/or pass the arguments for PHPUnit
+ as arguments to the script. Usage example:
+
+```~$ SYMFONY_VERSION=2.7.1 ./runTests.sh --filter testCustomFileNameProperty```
+
+**Note:** The script was prepared to run under Ubuntu and using Bash so it might need further validation for other OS.
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,6 @@
         "ocramius/proxy-manager": "~0.5"
     },
     "require-dev": {
-        "ext-mongo" : "*",
         "ext-sqlite3" : "*",
 
         "doctrine/mongodb-odm": "@dev",
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,9 @@
         "ocramius/proxy-manager": "~0.5"
     },
     "require-dev": {
+        "ext-mongo" : "*",
+        "ext-sqlite3" : "*",
+
         "doctrine/mongodb-odm": "@dev",
         "knplabs/knp-gaufrette-bundle": "*",
         "oneup/flysystem-bundle": "dev-master",
--- a/runTests.sh
+++ b/runTests.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+#
+
+#export SYMFONY_VERSION="~2.7.0"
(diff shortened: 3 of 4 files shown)


runTests.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+#
+
+#export SYMFONY_VERSION="~2.7.0"
+SUPPORTED_SYMFONY_VERSIONS=('~2.3.0' '~2.5.0' '~2.6.0' '~2.7.0' '2.8.x-dev')
+export DOCTRINE_VERSION=">=2.2.3,<2.5-dev"
+
+sudo apt-get -qq update
```
K-Phoen: This shouldn't be done here.

I like the general idea of this script but it should only be responsible for launching the tests (that includes playing with the `composer.json` file and optionally downloading composer).

runTests.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# Install/upgrade composer.phar
+curl -sS https://getcomposer.org/installer | php
+php composer.phar --version
+
+# Save a copy of the current composer.json file so we can restore it later
+cp composer.json composer.json.bck
+
+# Clear temp folder before the tests run
+sudo rm -rf /tmp/VichUploaderBundle/
```
K-Phoen: `sudo` isn't necessary (and can be scary when you launch the command without reading its sources :D)

K-Phoen:
Adding a note about this script and its usage in the [`CONTRIBUTING.md`](https://github.com/dustin10/VichUploaderBundle/blob/master/CONTRIBUTING.md#running-the-test-suite) guide would be nice :)

runTests.sh
Restless-ET: Are you referring to the update or the installs as well?

runTests.sh
Restless-ET: Indeed. :) I'll remove the ``sudo`` here.

runTests.sh
K-Phoen: The installs should be removed as well.

I don't think that this script should check that the packages are installed. The contributing documentation should mention them and composer could check for their presence (it already does it for php5-mongo for instance).

runTests.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+#
+
+SUPPORTED_SYMFONY_VERSIONS=('~2.3.0' '~2.5.0' '~2.6.0' '~2.7.0' '2.8.x-dev')
+export DOCTRINE_VERSION=">=2.2.3,<2.5-dev"
```
K-Phoen: why exporting this variable?

composer.json
```diff
@@ -25,6 +25,9 @@
         "ocramius/proxy-manager": "~0.5"
     },
     "require-dev": {
+        "ext-mongo" : "*",
```
K-Phoen: `ext-mongo` is required by `doctrine/mongodb-odm`, we do not need to require it ourselves.

CONTRIBUTING.md
```diff
@@ -32,7 +32,19 @@ Thank you!
...
-The test suite requires the `php5-mongo` extension to be installed.
+The test suite requires the `php5-mongo` and `php5-sqlite` extensions to be installed.
 
 PHPUnit should be installed by composer. Run the tests with the
 `./vendor/bin/phpunit` command.
+
+Alternatively you can use the `runTests.sh` bash script present at the project root.
+Default usage example:
+
+```~$ ./runTest.sh``` (This runs all tests with your current PHP version against all the support Symfony versions.)
```
K-Phoen: "*against all supported Symfony versions*"

runTests.sh
```diff
@@ -0,0 +1,38 @@
...
+  SUPPORTED_SYMFONY_VERSIONS=(${SYMFONY_VERSION})
+fi
+
+# Run the tests for each supported symfony version
+for sf_version in ${SUPPORTED_SYMFONY_VERSIONS[@]}; do
+  php composer.phar require --no-update "symfony/symfony":"${sf_version}"
+  php composer.phar require --no-update "doctrine/orm":"${DOCTRINE_VERSION}"
+
+  php composer.phar update symfony/symfony || { restore_composer && exit 1; }
+  php composer.phar install --prefer-source || { restore_composer && exit 1; }
```
K-Phoen: Why do we need to update AND install?

K-Phoen:
The script is starting to look really good :)

Making it a bit more verbose could be nice. Sample messages:

* *Using Symfony {VERSION} and Doctrine {VERSION}*
* *Installing dependencies*
* *Launching tests*

runTests.sh
Restless-ET: Blind copy/paste from the travis build which I forgot to adjust. :-/

composer.json
Restless-ET: I've checked the ``doctrine/mongodb-odm`` and [didn't found it](https://github.com/doctrine/mongodb-odm/blob/master/composer.json) that's why I've put it there explicitly.

But looks like it's required by ````doctrine/mongodb`` after all, which is a requirement of ``doctrine/mongodb-odm``. I'll be removing this line...

runTests.sh
Restless-ET: Again.. stupid me. :( I've started with install.. but then needed to update symfony due to the versions iteration and never switched to a "global update".

Restless-ET:
Suggestions and corrections applied. :)

K-Phoen:
Merged, thanks!

Issue #297, opened 2014-08-13
Upload is not performed if there is no change in the entity


K-Phoen:
When using doctrine, event listeners are not called if there have been no column changed. It's a problem because we precisely depend on these events to be fired to handle the uploads.

Right now, we use some boilerplate code to artificially trigger a change in the entity whenever a file is uploaded (by updating the `created_at` column for instance).

I'd like to find an other way to fix this issue.

rik43:
if we use update date in entity setter as in example

    public function setImageFile(File $image = null)
    {
        $this->imageFile = $image;

        if ($image) {
            // It is required that at least one field changes if you are using doctrine
            // otherwise the event listeners won't be called and the file is lost
            $this->updatedAt = new \DateTime('now');
        }
    }

this is execute in each time where entity is loaded. if after call $em->flush() updatedAt field will be updated in db - one query for each entity. 
it generate 1000+ queries in my db in some controllers.

garak:
@rik43 is must check if File is an UploadedFile. See https://github.com/dustin10/VichUploaderBundle/issues/8#issuecomment-3780172

rik43:
good idea, thanks

acantepie:
Is there a chance you fix this issue one day ?

Reproduced using :
- vich/uploader-bundle : 1.16.0
- symfony : 5.0.11
- php : 7.4.13

garak:
> Is there a chance you fix this issue one day ?

This is the wrong question.
The right question is: is there a chance **we** fix this issue one day?

acantepie:
Ok, but this issue was opened 5 years ago...

I you don't manage to fix it, maybe you should change this part of documentation :
```
    /**
     * If manually uploading a file (i.e. not using Symfony Form) ensure an instance
     * of 'UploadedFile' is injected into this setter to trigger the update. If this
     * bundle's configuration parameter 'inject_on_load' is set to 'true' this setter
     * must be able to accept an instance of 'File' as the bundle will inject one here
     * during Doctrine hydration.
     *
     * @param File|\Symfony\Component\HttpFoundation\File\UploadedFile|null $imageFile
     */
 ...
```

garak:
@acantepie it's already documented just 4 lines after

chancegarcia:
just ran into this issue.

bundle version: 1.19.1
symfony version: 5.4.8

[...]

Issue #8, opened 2012-02-02
The file is not updated if there are not other changes in the entity


ghost:
Say if i have a already existing record, and i'm trying to update it, in case if i ONLY upload file and no other entity properties get changed, the preUpdate event is not called, so the file is not uploaded.

Currently the workaround is to introduce some changes into the entity, before saving it.

PR #412, opened 2015-06-30
Update known_issues.md

Mondane:
Add 'instanceof'  check as written here: https://github.com/dustin10/VichUploaderBundle/issues/8#issuecomment-3780172 to avoid this issue https://github.com/dustin10/VichUploaderBundle/issues/297#issuecomment-86078245


K-Phoen:
Thanks!

PR #415, opened 2015-07-06
Clarify Doctrine documentation

kix:
It wasn't quite clear that keys in the YAML config should actually be the same as Doctrine entity fields. This PR clarifies this.

### DIFF
--- a/Resources/doc/mapping/yaml.md
+++ b/Resources/doc/mapping/yaml.md
@@ -4,10 +4,25 @@ YAML mappings
 You can choose to describe your entities in YAML. This bundle supports this
 format and comes with the following syntax to declare your uploadable fields:
 
+Here's how your entity could look like:
+
+```yaml
+Acme\DemoBundle\Entity\Product:
+    type: entity
+    # ...
+    fields:
+        # Product::$imageName will hold the reference to the upload
+        imageName:
+            type: string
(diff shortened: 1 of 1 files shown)


javiereguiluz:
@kix if the trouble you faced is that the doc doesn't explain that *"YAML config keys must be the same as Doctrine entity fields"*, don't you think that the new documentation should explicitly say so?

kix:
@javiereguiluz, hm, right, good idea; will update this PR.

Resources/doc/mapping/yaml.md
```diff
@@ -4,10 +4,25 @@ YAML mappings
...
+            type: string
+            column: image_name
+```
+
+And here's how you should configure the uploader:
+
 ```yaml
 # src/Acme/DemoBundle/Resources/config/vich_uploader/Product.yml
 Acme\DemoBundle\Entity\Product:
-    image:
```
K-Phoen: Hum, no. The documentation is correct, this key is the right one.

K-Phoen:
Yaml config keys reference the attribute in which uploaded file objects will be stored (ie: `UploadedFile` instances).
The entity field used to store the file name is defined by the `filename_property` Yaml property.

K-Phoen:
Closing this issue.

PR #426, opened 2015-08-04
Remove name parameter from field mapping

bocharsky-bw:
The `name` could override `doctrine.orm.naming_strategy`.
Other developers could use different naming strategy so would be better to avoid it in example in favor to easy copy/paste.


K-Phoen:
Thanks :+1:

PR #427, opened 2015-08-04
Remove element names in @var annotation and use namespace

bocharsky-bw:
I think for simlicity we could avoid element names. The element name in `@var` is optional and not required. If anybody copy/paste this code from examples he most probably change the names so he should change it in few places (in code and annotation).


K-Phoen:
I guess you're right, thanks for the PR :)

Issue #389, opened 2015-02-25
Force downloading uploaded images


bocharsky-bw:
Uploaded images opening in a browser tab when click on **Download** link.

Can we force downloading for any files?

K-Phoen:
Images are directly served by the web server so I'd say no.

bocharsky-bw:
What about if we make download link via controller? We can add forcing headers in this case

K-Phoen:
This would work (despite being a bit heavy).

If you want to submit a PR I'd be happy to review it :)

bocharsky-bw:
Ok, I think about that :) Maybe later I'll have a time

K-Phoen:
I kind of forgot about this issue, sorry!

There is - at last - some documentation about the download helper. You can find it here: https://github.com/dustin10/VichUploaderBundle/blob/master/Resources/doc/downloads/serving_files_with_a_controller.md

Sorry again.

bocharsky-bw:
Oh, nice! Thanks for docs!

PR #430, opened 2015-08-05
Force download of uploaded files

bocharsky-bw:
This PR based on https://github.com/dustin10/VichUploaderBundle/issues/389

### DIFF
--- a/Controller/VichController.php
+++ b/Controller/VichController.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace Vich\UploaderBundle\Controller;
+
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\BinaryFileResponse;
+
+/**
+ * Class VichController
+ *
+ * @author Victor Bocharsky <bocharsky.bw@gmail.com>
+ */
+class VichController extends Controller
+{
+    /**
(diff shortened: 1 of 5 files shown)


bocharsky-bw:
We need to import routing.xml file in `app/config/routing.yml`:
``` yaml
_vich_uploader:
    resource: "@VichUploaderBundle/Resources/config/routing.xml"
    prefix: /_vich
```

I'll add it to the docs config section if it's OK.

K-Phoen:
I only did a quick review of this PR but I already can tell that I won't merge it as is.
You consider that all uploaded files are stored in the `web` directory but it's just wrong. They can be in a directory that's not accessible by the webserver or they can even be in another server (Amazon S3 for instance).

For these cases, I already wrote a [download handler](https://github.com/dustin10/VichUploaderBundle/blob/5d7fc2c9b36a450143fd5429e540112435c6be60/Handler/DownloadHandler.php) that essentially creates a `StreamResponse` to the file that we want to download, even if it is stored in a remote server.

We probably need to mention this handler in the documentation, maybe provide a better integration with the form type or something else (I'm not sure about that yet).

bocharsky-bw:
Yeah, looks great! And integration with a form types will be cool!
I'm also very interesting in usage examples of it. When will you plan to add some docs?

PR #431, opened 2015-08-05
Add PHP 7.0 support by Travis

bocharsky-bw:
What about also to add `hhvm`? We could ignore failures if there are any problems with it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,11 @@ php:
   - 5.4
   - 5.5
   - 5.6
+  - 7.0
+
+matrix:
+  allow_failures:
+    - php: 7.0
 
 sudo: false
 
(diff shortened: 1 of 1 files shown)


K-Phoen:
The tests for PHP 7.0 are not even launched: https://travis-ci.org/dustin10/VichUploaderBundle/jobs/74202141

Looks like the `.travis.yml` file needs a few changes.

bocharsky-bw:
Yeah, I don't understand yet what is the problem.
I think `7.0.0` is coming so we should fix it.

K-Phoen:
It fails because we try to disable xdebug (see instruction `phpenv config-rm xdebug.ini`) but it's not enabled for PHP 7.

bocharsky-bw:
Thanks, I fixed it! But seems there is another error with `mongodb`

K-Phoen:
Yep, the bundle has a dependency to the php-mongo extension for some tests. I guess that we should be able to skip these tests if the extension isn't installed, I'll try to do that soon and I'll merge this PR after.

One last thing: can you squash your commits please?

Thanks again for your work :)

bocharsky-bw:
Yes, I squashed all my commits and did rebase with master

K-Phoen:
Merged in #436, thanks for your contribution!

bocharsky-bw:
Thanks for improvements! )

PR #432, opened 2015-08-05
Self update Composer before install packages

bocharsky-bw:
Travis use outdated version of Composer, better to update it first


K-Phoen:
Out of curiosity: why is it necessary?

bocharsky-bw:
We had some problem with outdated `Composer` in https://github.com/symfony/symfony-demo/pull/114 .
Anyway I think it allows to avoid akward errors in future :-)

K-Phoen:
Okay then :+1:

PR #433, opened 2015-08-05
Make properties as private in docs examples

bocharsky-bw:
I think better to make properties `private` and use other modifiers only if you need.
Another one improvement for convenient copy/paste.


K-Phoen:
Thanks again :D

Issue #447, opened 2015-09-08
Using an instance of "Twig_Function_Method" for function "vich_uploader_asset" is deprecated. Use Twig_SimpleFunction instead


gpapin:
On the UploaderExtension you are using the deprecated way of defining extension in twig that will get removed in twig 2.0

This gives you this warning on the profiler :
```DEPRECATED - Using an instance of "Twig_Function_Method" for function "vich_uploader_asset" is deprecated. Use Twig_SimpleFunction instead.```

```php
    /**
     * Returns a list of twig functions.
     *
     * @return array An array
     */
    public function getFunctions()
    {
        $names = array(
            'vich_uploader_asset' => 'asset'
        );

        $funcs = array();
        foreach ($names as $twig => $local) {
            $funcs[$twig] = new \Twig_Function_Method($this, $local);
        }

        return $funcs;
    }
```

mvhirsch:
I just provided a PR for this. See #448

gpapin:
Thanks guys

PR #448, opened 2015-09-09
remove deprecation warning while creating twig function

mvhirsch:
Providing PR for #447

### DIFF
--- a/Tests/Twig/Extension/UploaderExtensionTest.php
+++ b/Tests/Twig/Extension/UploaderExtensionTest.php
@@ -28,9 +28,9 @@ class UploaderExtensionTest extends \PHPUnit_Framework_TestCase
     public function testAssetIsRegistered()
     {
         $functions = $this->extension->getFunctions();
-        $expectedFunction = new \Twig_SimpleFunction('vich_uploader_asset', array($this->extension, 'asset'));
 
-        $this->assertEquals(array($expectedFunction), $functions);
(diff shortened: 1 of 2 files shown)


Tests/Twig/Extension/UploaderExtensionTest.php
```diff
@@ -27,7 +27,10 @@ public function testGetName()
 
     public function testAssetIsRegistered()
     {
-        $this->assertArrayHasKey('vich_uploader_asset', $this->extension->getFunctions());
+        $functions = $this->extension->getFunctions();
+        $expectedFunction = new \Twig_SimpleFunction('vich_uploader_asset', array($this->extension, 'asset'));
+
+        $this->assertEquals(array($expectedFunction), $functions);
```
K-Phoen: This equality test is meaningless IMHO.

We should test that the extension defines exactly one function and that this function is named `vich_uploader_asset`.
This would make the test-case look like this:

```php
$this->assertCount(1, $functions);
$this->assertEquals('vich_uploader_asset', $functions[0]->getName());
```

K-Phoen:
I'll merge this PR once my comment is addressed.

Thanks for submitting it! :+1:

mvhirsch:
Sorry, I was on vacation last week, so here we go :-)

K-Phoen:
Perfect, thanks!

PR #451, opened 2015-09-18
Fix docs

Miliooo:
There is no clear:cache command, the command is cache:clear


K-Phoen:
You're right, nice catch! :+1:

Issue #453, opened 2015-10-01
DEPRECATED - FormTypeInterface::setDefaultOptions()


patie:
DEPRECATED - Vich\UploaderBundle\Form\Type\VichFileType: The FormTypeInterface::setDefaultOptions() method is deprecated since version 2.7 and will be removed in 3.0. Use configureOptions() instead. This method will be added to the FormTypeInterface with Symfony 3.0.

K-Phoen:
If you want to write a PR for that, I'd be happy to review it :)

belendel:
Made a PR for it.

K-Phoen:
Should be fixed in master.

PR #455, opened 2015-10-13
Remove deprecated methods from form type

belendel:
Fix for issue #453


belendel:
Should we work with a different branch for new version of symfony?

K-Phoen:
Fixed by another PR.

PR #459, opened 2015-10-18
Make resolving of paths optional for delete link in VichFileType

MrNicodemuz:
My staging server does not have all the files always on the server itself, thus making most of my forms not render at all.

### DIFF
--- a/Form/Type/VichFileType.php
+++ b/Form/Type/VichFileType.php
@@ -32,6 +32,7 @@ class VichFileType extends AbstractType
     {
         $resolver->setDefaults(array(
             'allow_delete'  => true,
+            'delete_assert_file_exists' => true,
             'download_link' => true,
         ));
     }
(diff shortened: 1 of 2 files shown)


MrNicodemuz:
@dustin10, any feedback? :)

K-Phoen:
Introducing a new option in the form type to solve a case of missing fixtures in a staging environment seems a bit overkill.

`resolvePath` is used to determine if a file has been uploaded but it checks that the file is present in the filesystem, hence your issue. Another solution would be to use `resolveUri` as it does not check the presence of the file (if I remember correctly).

MrNicodemuz:
@K-Phoen: Thanks for your valuable feedback. I have updated the pull request to use ```resolveUri``` instead.

K-Phoen:
Thanks!

I'd like to test a few things with this PR before merging it. Don't worry if I take a few days to merge this PR, I don't have a lot of free time right now :)

MrNicodemuz:
Bump! :)

K-Phoen:
Sorry for the delay, I quickly tested it but I forgot to merge the PR :(

Thanks for your contribution :+1:

PR #464, opened 2015-10-27
Deprecated issue fixed #453 

### LINKED ISSUE #453: DEPRECATED - FormTypeInterface::setDefaultOptions()

bobod:
composer.json updated because of PHPUnit test. It depends on the added packages.
#453

### DIFF
--- a/Form/Type/VichFileType.php
+++ b/Form/Type/VichFileType.php
@@ -8,6 +8,7 @@ use Symfony\Component\Form\FormEvent;
 use Symfony\Component\Form\FormEvents;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\Form\FormView;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\OptionsResolver\OptionsResolverInterface;
 use Symfony\Component\Translation\TranslatorInterface;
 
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -18,7 +18,7 @@
     },
     "require": {
         "php": ">=5.4",
-        "symfony/framework-bundle": "~2.3",
+        "symfony/framework-bundle": "~2.7.5",
```
K-Phoen: Hum, it means that all Symfony versions prior to 2.7.5 would not be supported by this bundle. That's a problem, we do want to support them.

bobod: Yes, it's just a mistake, because I had to add some packages to the composer.json in local environment in order to run unit tests.

K-Phoen:
This PR looks good to me. There is no BC break and I trust you when you say that it solves the deprecation warning.

Could you [squash your commits](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html)? I'll merge the PR once it's done.

bobod:
This was the first time when I squashed my commits, but I think I've done with it.

K-Phoen:
Merged by [this commit](https://github.com/dustin10/VichUploaderBundle/commit/27888d5aa2e6676fbfb565dd47eac8c538735400).

Thanks :+1:

PR #465, opened 2015-10-27
UniqidNamer: Change visibility to protected

foaly-nr1:
When using a custom file namer that extends UniqidNamer, getExtension() can only be overridden when visibility is not private. I think it's good practice to allow extending classes when publishing bundles for common use.

### DIFF
--- a/Naming/Polyfill/FileExtensionTrait.php
+++ b/Naming/Polyfill/FileExtensionTrait.php
@@ -13,7 +13,7 @@ trait FileExtensionTrait
      *
      * @return string|null
      */
-    private function getExtension(UploadedFile $file)
+    protected function getExtension(UploadedFile $file)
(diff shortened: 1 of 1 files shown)


K-Phoen:
The namers are not meant to be extended. I could have defined them as final classes but I didn't want to add a bc-break just for that.
If you need another naming strategy, then you need another class. the `getExtension()` method is defined in a trait so you'll still be able to use it.

Generally speaking, I think that it's safer to define as private as much methods as possible. These methods, when re-used through extension by users of the bundle/library become a *de facto* "public API" that I have to maintain and with which I can not break compatibility.

I'd be glad to read your arguments, but for now I think that I won't merge this PR.

foaly-nr1:
Maintenance is a good point.

PR #466, opened 2015-10-27
Compute change set after updating all fields

foaly-nr1:
In an entity with multiple uploadable fields, is there a reason to compute the change set before all fields have been updated?


K-Phoen:
Thanks!

PR #470, opened 2015-11-09
Improve Travis build speed by reducing amount of jobs

WouterJ:
This also tests the claim that this bundle supports Symfony 3.

You don't need to test every version of Symfony against every PHP version. This PR changes the logic so only 2.7 is tested against every PHP version and all other versions are tested against the latest stable PHP version (5.6). This means instead of `supported_php_versions * supported_symfony_versions`, you now have `supported_php_versions + supported_symfony_versions` jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,24 +1,36 @@
 language: php
 
 php:
-  - 5.4
-  - 5.5
-  - 5.6
-  - 7.0
+    - 5.4
+    - 5.5
+    - 5.6
+    - 7.0
 
 sudo: false
 
-env:
-    - SYMFONY_VERSION="~2.3.0"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,24 +1,36 @@
...
+    - 5.4
+    - 5.5
+    - 5.6
+    - 7.0
 
 sudo: false
 
+cache:
+    directories:
+      - $HOME/.composer/cache
```
WouterJ: caching composer will make downloading a lot faster (as dists are already downloaded).

K-Phoen:
I fixed the mongodb-related issues and merged this PR.

Thanks :+1:

PR #474, opened 2015-11-24
Fix DownloadHandler

jvasseur:
The result of stream_copy_to_stream is the number of byte copied and shouldn't be echoed.


K-Phoen:
Looks good, thanks!

PR #475, opened 2015-11-26
Catch and bury exceptions when injecting files

CodingNinja:
Prevent file missing etc from causing exceptions during injection

### DIFF
--- a/Injector/FileInjector.php
+++ b/Injector/FileInjector.php
@@ -34,10 +34,14 @@ class FileInjector implements FileInjectorInterface
      */
     public function injectFile($obj, PropertyMapping $mapping)
     {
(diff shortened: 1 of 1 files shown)


K-Phoen:
Hi!

What's your use-case for this? This check is here to ensure consistency, why do you want to remove it? (silently injecting a file that does not exist does not make much sense to me)

CodingNinja:
Hi mate!

This was a quick hack that prevents issues when using FlySystem with a file that doesn't exist (Using a production DB backup for example). 

I think it's valid as the method just injects the file but if one has been uploaded but if the underlying file provider throws an exception, there must be a problem which will prevent the file being read anyway and rather than throw an exception which will most likely interrupt the request, it is silently caught and dropped.

Maybe this is better off being a config option for the injector?

PR #476, opened 2015-11-29
Add `fresh/vich-uploader-serialization-bundle` to suggest section of composer.json

fre5h:
https://github.com/fre5h/VichUploaderSerializationBundle provides an easy integration with JMSSerializerBundle by using two additional annotations. Would be nice to include this bundle into suggestions of VichUploaderBundle.


K-Phoen:
Sorry but I'm not willing to add an almost unknown bundle the the suggestions of this project (which is already quite long…)

PR #479, opened 2015-12-03
[WIP] travis allowed to fail with symfony 3

tgabi333:


K-Phoen:
Removing the "@dev" annotations is a good idea and so is testing against PHP7. But considering that the tests are allowed to fail against Symfony 3.0 might not be. The tests are supposed to pass and if they don't, I want them to be red (even if they fail because of a dependency resolution problem).

Could you remove symfony 3 from the list of allowed failures? I'll merge this PR once it's done.

tgabi333:
allow failures removed :+1:

K-Phoen:
Thanks! :+1:

PR #480, opened 2015-12-03
remove test namespace from autoloading

tgabi333:
It removes 3 files from autoloading, not much, but a good start to clean up the optimized autoloader maps.

A lot of test files remains in the loader path because the root directory is the autoloading psr-4 root directory too. It should be separated to another dir.


K-Phoen:
Thanks!

PR #498, opened 2016-01-12
[Code style] Import all FQCNs when possible

Pierstoval:
It might be cool to simplify all class imports.

I noticed there was a bug in some phpdoc in v0.14.0 because my IDE wasn't auto-completing correctly.

### DIFF
--- a/Event/Event.php
+++ b/Event/Event.php
@@ -35,7 +35,7 @@ class Event extends BaseEvent
     /**
      * Accessor to the mapping used to manipulate the object.
      *
-     * @return \Vich\UploaderBundle\Mapping\PropertyMapping
+     * @return PropertyMapping
      */
     public function getMapping()
     {
--- a/Handler/AbstractHandler.php
+++ b/Handler/AbstractHandler.php
@@ -12,18 +12,18 @@ use Vich\UploaderBundle\Storage\StorageInterface;
 abstract class AbstractHandler
 {
     /**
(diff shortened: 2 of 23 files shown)


K-Phoen:
This PR breaks the tests ;)

Besides, I'm not sure I want to add unused `use` statements just to make IDE happy.
Replacing a FQCN by a simple class name where it's possible is definitely fine though.

Pierstoval:
I updated the PR to fix the tests, it is green now

JHGitty:
:+1: +1 for this code style.

robfrawley:
Please merge this... Using FQCN everywhere in the docs is a recipe for disaster as it introduces repetitious strings throughout the code base that must all be updated when something is moved/renamed/etc. It's much easier to update the use statements than to update the 10, 20, 100, or more repeated docblocks.

Moreover, there is no overhead on the engine as zend does nothing with a use statement until it comes across a piece of code that actually references the class.

In my opinion this has very little to do with IDEs --- any modern IDE can happily perform code completion regardless of whether you utilize use statements and short names in docs blocks or FQCN in doc blocks. To me it's just a matter of clean and easy to maintain source... But that's just me.

Pierstoval:
I've rebased and I fixed one small minor code style issue in a doc file :stuck_out_tongue:

Pierstoval:
👏 Thanks

PR #499, opened 2016-01-15
Add error_bubbliing option

### LINKED ISSUE #460: form_errors with form theme
Hello i have an issue using your form theme

I'm using an assert on my imageFile : * @Assert\Image(maxSize="300k", mimeTypes={"image/jpeg", "image/png"}, allowLandscape=true, allowPortrait=false)

But when I validate the form, the imageFile errors are considered as global error and not as field error :/

It's not a global error when I'm not using your form theme.

Thank you !

garak:
Fix #460


K-Phoen:
Thanks! :)

PR #506, opened 2016-01-20
Add return value to setter in example

JHGitty:
This should be the same behaviour as the Doctrine entity generator would create.


K-Phoen:
Thanks :)

Issue #508, opened 2016-01-20
Travis and runTests.sh should test the same


JHGitty:
Thanks to @geekdevs I found this issue. (https://github.com/dustin10/VichUploaderBundle/issues/489#issuecomment-171189911)

Currently in Travis we do **not** test on Symfony versions ``2.5.0`` and ``2.6.0``. But in ``runTests.sh`` we test against these Symfony versions. Version ``2.4.0`` is missing in both tests.

There are 2 possible solutions:

1. We do no longer test against these Symfony versions because they are End-of-Life since 07/2015 and 01/2016. http://symfony.com/doc/current/contributing/community/releases.html#schedule

2. We add ``2.5.0`` and ``2.6.0`` to Travis and ``2.4.0`` to both.

What do you think?

JHGitty:
@K-Phoen What do you think should we do?

K-Phoen:
We are supposed to support `~2.3|~3.0`, or at least that's what our `composer.json` says. So in my opinion we should fix the bundle for the broken versions and add at least one entry for `~2.4`, `~2.5` and `~2.6` in the `.travis.yml` file.

I know that these versions are not supported anymore by the Symfony core team, but ideally I'd like this bundle to work seamlessly from Symfony ~2.3 to ~3.0.

Would you like to take the lead on this issue?

JHGitty:
@K-Phoen I agree with you. The ``composer.json`` file is the most important reason why we should check against all Symfony versions.

I'll create a Pull Request for Travis. Maybe we will find some BC breaks / broken tests. Hopefully Travis will run the tests successfully.

geekdevs:
Note that the problem is not only that `runTests.sh` tests different versions than travis, but also that `runTests.sh` fails on the versions which travis does not.

garak:
:-1:  for supporting unsupported versions of Symfony.
You could find yourself into a situation where your software is failing for a problem caused by Symfony and you cannot make Symfony fix that problem.
So you're not really supporting such versions, in the end.

K-Phoen:
I don't know yet what I want to do for the 1.0 release.

The `composer.json` files indicates that the bundle supports from Symfony ~2.3 to ~3.0. We are basically saying that the bundle works with Symfony 2.3, 2.4, 2.5, … up to 3.0. So including these versions in our test matrix kind of make sense (even if they are not maintained anymore).
That induces a longer run-time for our test builds and more maintenance-related work.

[...]

PR #509, opened 2016-01-20
Add missing Symfony versions to tests

JHGitty:


.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: php
 
 php:
-    - 5.5
```
K-Phoen: Why removing PHP 5.5 ?

.travis.yml
```diff
@@ -16,9 +15,15 @@ env:
 
 matrix:
     include:
-        - php: 5.5
```
K-Phoen: Same here, we do want to test on both the lowest supported Symfony and PHP versions.

.travis.yml
JHGitty: We could check against PHP 5.5 **and** PHP 5.6. But why should we check only against PHP 5.5 on Symfony 2.3?

.travis.yml
JHGitty: @K-Phoen Thanks! Discussion see other comment.

JHGitty:
@K-Phoen
Should we remove this line? As I understand the matrix includes already contain the Symfony Version. 
Or is it important to have this line still there?

```yml
env:
    - SYMFONY_VERSION="~2.8.0"
```

.travis.yml
K-Phoen: See [this PR](https://github.com/dustin10/VichUploaderBundle/pull/470) :)

JHGitty: I understand. Thanks. I'll fix it.

K-Phoen: Awesome, thanks :+1:

tgabi333:
Symfony 2.4, 2.5 and 2.6 are not supported anymore, so there is no need for testing. 

http://symfony.com/doc/current/contributing/community/releases.html

K-Phoen:
@tgabi333 I know, but this bundle claims to support Symfony ~2.3 so I it make sense to test 2.4, 2.5 and 2.6 ;)

K-Phoen:
Closing this since it's solved in master (see #508)

PR #510, opened 2016-01-20
Remove PHP requirement from composer.json

JHGitty:
We should allow PHP 5.3 and PHP 5.4. The PHP requirement check is already done by the Symfony ``symfony/framework-bundle``.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,6 @@
         }
     },
     "require": {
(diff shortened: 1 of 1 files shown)


K-Phoen:
Nope, we are stricter than the framework and it's done on purpose ;)

PR #522, opened 2016-02-05
Use double slashes in twig example

benoliver999:
Otherwise, twig just crunches it all together, eg `FooBundleEntityProduct`, and throws a `Class not found` Reflection Exception.


K-Phoen:
Thanks :)

PR #541, opened 2016-03-07
Additional ocramius/proxy-manager version (1.0 and 2.0)

dawiddominiak:
There are some problems with using older version of "ocramius/proxy-manager" in PHP 7.

Argument parameter types such as string, float, int, bool are not converted in the proper way during the caching process. Declarations of cached/proxy methods are not compatible with their's parent equivalent.

On the other hand 2nd version of "ocramius/proxy-manager" is just stable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,10 +28,11 @@ matrix:
           env: SYMFONY_VERSION="~3.0.0"
 
 before_install:
+    - echo "extension = mongodb.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
     - composer self-update
     - phpenv config-rm xdebug.ini || true
     - composer require --no-update symfony/symfony:${SYMFONY_VERSION}
 
-install: composer update --prefer-dist --ignore-platform-reqs
+install: composer update --prefer-dist
 
 script: vendor/bin/phpunit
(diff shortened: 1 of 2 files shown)


K-Phoen:
Requiring ~2.0 is not an option as this version is only compatible with PHP ~7.0.
This bundle should allow both ~1.0 and ~2.0. Can you update this PR ? :)

dawiddominiak:
OK, I've updated that but there is a problem in travis command.

`composer update` is running with option `--ignore-platform-reqs`.
[Documentation](https://getcomposer.org/doc/03-cli.md#install)

Composer always try to install newer version of `ocramius/proxy-manager` even with older php version. Everything should work without that option.

dawiddominiak:
OK, it's not working proper.

There is missing `ext-mongo` in your travis environment. I've tried to change `doctrine/mongodb-odm` version to not use `ext-mongo` directly, but it's still required by `doctrine/mongodb`.

Compatibility with PHP7 in `doctrine/mongodb` is hold by the following issue: https://github.com/doctrine/mongodb/issues/239 . 

I'll try keep my fork till issue above will be done.

jrobeson:
bitten by this one too now.

K-Phoen:
Should be fixed in master and released as 1.0.1 :)

PR #545, opened 2016-03-14
Spanish translation

abdielcs:
Added Spanish translation.

### DIFF
--- a/Resources/translations/VichUploaderBundle.es.yml
+++ b/Resources/translations/VichUploaderBundle.es.yml
@@ -1,0 +1,3 @@
+download: Descargar
+
+form.label.delete: Eliminar?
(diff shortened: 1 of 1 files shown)


Resources/translations/VichUploaderBundle.es.yml
```diff
@@ -0,0 +1,3 @@
+download: Descargar
+
+form.label.delete: Eliminar?
```
javiereguiluz: It should be: `¿Eliminar?`

abdielcs:
You're right, I already added the commit.

Resources/translations/VichUploaderBundle.es.yml
abdielcs: You're right, I already added the commit.

K-Phoen:
Thanks! :+1:

javiereguiluz:
@abdielcs thanks for the translation!

PR #548, opened 2016-03-16
add parent to types

garak:
This should fix the parent issue referred in #401 with compatibility with Symfony 3.
Unfortunately I wasn't able to run tests, since composer doesn't install. I hope everything is fine.

### DIFF
--- a/Form/Type/VichFileType.php
+++ b/Form/Type/VichFileType.php
@@ -133,4 +133,12 @@ class VichFileType extends AbstractType
 
         return $shortType;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getParent()
+    {
+        return $this->getFieldType('file');
+    }
 }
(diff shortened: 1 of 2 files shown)


garak:
Unfortunately, it looks like this is just not working.
Symfony is complaining that you cannot use `add` in `buildForm` without setting `compound` to `true`. Setting `compound` to `true` breaks everything: `The form's view data is expected to be an instance of class Symfony\Component\HttpFoundation\File\File, but is a(n) array`

K-Phoen:
So, does the problem described in #401 really exists and if so, is it blocker?
If it's the case, our two form types might need a bit of refactoring.

WDYT?

garak:
Well, I'm using VichFileType on a situation really similar to the one described in #401 and I never had any problem.
My case is: EntityA in oneToMany with EntityB with cascade persist. EntityB is VichUploadable and its form uses a VichFileType, the form of EntityA is saved without problem

Dragnucs:
@garak I am using this bundle with Symfony 3.0.3, Installed with provided documentation and it is working fine. I cannot reproduce your problem.

K-Phoen:
Given that there is no other report of this issue, I think we can ignore it and release 1.0.0 today :)

garak:
Great! I'm closing this

Issue #163, opened 2013-10-30
Failing to set metadata before uploading the file


pmmaga:
I'm using a combination of VichUploader with Gaufrette to upload files to an Azure Blob Storage.

My issue is that when I try to make an upload, since the AzureBlobStorage implements the MetadataSupporter interface, the GaufretteStorage will try to set the metadata for the file. This occurs before the file is uploaded so I get a 404 when I try to upload the file:
"Failed to set metadata for blob "xxxxxx" in container "xxxxxx": The specified blob does not exist. (BlobNotFound)."

If I remove the subscription for the MetadataSupporter interface or if I move the setting of the metadata after the file is uploaded, everything works properly. Has anyone had a similar experience? Am I looking for the problem in a wrong perspective?

PR #556, opened 2016-04-11
make gaufrette's metadata work

dswbx:
make gaufrette's metadata work by creating Gaufrette\File and set metadata there

### DIFF
--- a/Storage/GaufretteStorage.php
+++ b/Storage/GaufretteStorage.php
@@ -4,6 +4,7 @@ namespace Vich\UploaderBundle\Storage;
 
 use Gaufrette\Adapter\MetadataSupporter;
 use Gaufrette\Exception\FileNotFound;
+use Gaufrette\File;
 use Knp\Bundle\GaufretteBundle\FilesystemMap;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
 use Vich\UploaderBundle\Mapping\PropertyMapping;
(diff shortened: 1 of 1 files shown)


garak:
@RSSfeed can you tell something more about this PR?

RSSfeed:
It's been a long time, but as I remember, without this changes, files stored via gaufrette in amazon s3 won't get the correct metadata. E.g. a jpg got mime-type (as far as I remember) "application/octet-stream" instead of "image/jpeg". Without having the correct mime-type the browser won't behave as expected.

With my pull request changes in place the metadata is being added correctly.

Koc:
@RSSfeed can you add test that confirm fix and prevent regression in future?

garak:
@RSSfeed did you take a look to #163?

OJezu:
What is preventing this PR from being merged?

garak:
* This branch has conflicts that must be resolved
* Missing tests
* Awaiting feedback about another possible related issue

OJezu:
This does not fix #163, as when setting metadata in Gaufrette File, Gaufrette first tries to save metadata, and then the file itself. See KnpLabs/Gaufrette#537

Upon closer inspection I don't know if this PR actually changes anything, as 
```
$gaufretteFile = new File($path, $filesystem);
$gaufretteFile->setContent(file_get_contents($file->getPathname()), array('contentType' => $file->getMimeType()));
```

Underneath calls the same `$filesystem->getAdapter()->setMetadata` and `$filesystem->getAdapter()->write`. Workaround for Gaufrette issues would be to write file first, and then setMetadata. If KnpLabs/Gaufrette#537 is fixed, then 
```
$gaufretteFile = new File($path, $filesystem);
$gaufretteFile->setContent(file_get_contents($file->getPathname()), array('contentType' => $file->getMimeType());
```
Could be done no matter if adapter supports metadata or not, and I guess that would be the best solution. It should be up to Gaufrette to save metadata along with file in most atomic way.

garak:
So, what we can do here is waiting for Gaufrette feedback, then close this PR (that looks too old) and open a new one.

OJezu:
Without Gaufrette feedback, Uploader can save file first, then metadata at risk of file being saved without metadata. Or retry saving metadata after file upload, if it failed before. Those are options to consider, if Gaufrette is not fixed.

I looked into fixing Gaufrette myself, but after seeing code I put it off - there is no way to do it there without risking some (internal) BC changes. I will see how KnpLabs/Gaufrette#537 goes.

garak:
Closing because it's too old and without any feedback

PR #557, opened 2016-04-13
Force 'has-error' when invalid

garak:
Before this change, an invalid image/file field was missing the `has-error` class, resulting in missing red coloring of label and error

before:
before

after:
after


K-Phoen:
Thanks :+1:

PR #559, opened 2016-04-15
Hint to usual file overriding issue.

Dragnucs:


K-Phoen:
Good idea, thanks :+1:

PR #574, opened 2016-05-24
[Composer] Refine dependencies

egeloen:
Hey!

This PR refines the composer dependencies in order to be able to run phpunit tests locally just after running a `composer install` (a bunch of dependencies were missing if we use Symfony 3).

It also refines dependencies in order to make tests green even if we install with the lowest deps: `composer install --prefer-lowest --prefer-stable`.

Since travis is ran with the full `symfony/symfony`, this has not been detected previously.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,33 +18,45 @@
     },
     "require": {
         "php": "~5.5|~7.0",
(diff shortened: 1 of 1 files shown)


K-Phoen:
Thanks :+1:

PR #575, opened 2016-05-26
Add custom storage support

### LINKED ISSUE #573: Custom storage support
Hey!

I'm currently implementing a custom storage which will give me support for [Cloudinary](http://cloudinary.com/) but since 31b2900e2a0c2159e40ac48ad00f83a598c6a84d we are not able to put a custom service name as storage in the configuration (See [here](https://github.com/dustin10/VichUploaderBundle/blob/master/DependencyInjection/Configuration.php#L17) and [here](https://github.com/dustin10/VichUploaderBundle/blob/master/DependencyInjection/Configuration.php#L49-L52)).

I'm not sure if right now, it is supported to add custom storage since the configuration really restrict the allowed values. I have some solutions to propose but none which would not break BC...

What about dropping this validation in the configuration?

egeloen:
This PR fixes #573 by allowing to define a custom storage service name in the configuration. 

To to not break the BC, I have chosen to prefix the service name with an `@` in order to inform the DI extension we are using a custom storage service. I think the `@` char is the best option since it is also used in yaml service definition :)

WDYT?

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -58,8 +58,10 @@ class Configuration implements ConfigurationInterface
                         ->then(function ($v) { return strtolower($v); })
                     ->end()
                     ->validate()
-                        ->ifNotInArray($this->supportedStorages)
-                        ->thenInvalid('The storage %s is not supported. Please choose one of ' . implode(', ', $this->supportedStorages))
+                        ->ifTrue(function ($storage) {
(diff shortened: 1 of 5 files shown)


K-Phoen:
After a quick review, I don't see any problem. I'll take more time later in the day to do a real review and hopefully merge it right after :)

:+1: for the tests and the doc

K-Phoen:
Merged, thanks for this PR :)

PR #577, opened 2016-05-27
Fix metadata changes detection

egeloen:
Hey!

Since I'm currently playing with this bundle, I submit a bugfix about metadata changes detection. Basically, the JMS metadata factory was always in no debug mode meaning once the metadata were loaded, no more checks were done on them.

This PR fixes it by propagating the `kernel.debug` flag into the metadata factory and then, bring back the metadata changes detection :)


K-Phoen:
Nice one, thanks! :+1:

PR #591, opened 2016-07-13
Add support for propel-events from GlorpenPropelBundle.

glorpen:
I've bumped into problem when creating custom driver - when using `propel` orm option,
`getUploadableClasses` and `getUploadableFields` are called when Container is not yet built,
and because of it some base services are unavailable - eg. `kernel`.

This PR adds support for GlorpenPropelBundle as another "orm": `propel_ge`, when used allows to skip
reading data from not-yet-built Container in `RegisterPropelModelsPass`.
Going further, event handling looks more like in Doctrine part of this bundle and thus enables a way
to write custom drivers when using Propel.

### DIFF
--- a/Adapter/Propel/PropelORMGeAdapter.php
+++ b/Adapter/Propel/PropelORMGeAdapter.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Vich\UploaderBundle\Adapter\Propel;
+
+use Vich\UploaderBundle\Adapter\AdapterInterface;
+
+/**
+ * Propel adapter for GlorpenPropelBundle.
+ *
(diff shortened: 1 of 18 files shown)


glorpen:
I've noticed that I didn't update readme, I'll fix that - but before that I would like to know if this feature is acceptable & in right places :)

garak:
Can you fix conflicts?

glorpen:
I think we can just close this PR.
Legacy Propel is more likely to be accompanied with EOLed Symfony 2+ ... and to make this PR viable there would be changes in VichUploaderBundle codebase to support EOL Symfony, which, I think, is not worth it :)

PR #607, opened 2016-08-08
Defer RegisterPropelModelsPass

cedricziel:
The changeset was necessary after a change in symfony master.

The service 'cache.annotations', which is resolved when the metadata
reader is received from the container in the Pass, is not available
at that point in the container compilation process.

This changeset defers the PropelModelsPass until a point in time, where the
container is fully accessible.

Ref: symfony/symfony#19205

### DIFF
--- a/VichUploaderBundle.php
+++ b/VichUploaderBundle.php
@@ -2,6 +2,7 @@
 
 namespace Vich\UploaderBundle;
 
+use Symfony\Component\DependencyInjection\Compiler\PassConfig;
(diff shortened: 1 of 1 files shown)


cedricziel:
Sure thing. I added it, because the build kept failing due to memory_limit. Can't so it at the moment, but will do and ping.

cedricziel:
I reverted the change on `.travis.yml` as requested.

garak:
Thank you

cedricziel:
Thank you for merging :)

PR #617, opened 2016-09-01
[WIP] Use preFlush event introduced in Doctrine 2.2 (closes #89)

### LINKED ISSUE #89: How modify a image ?
Hello.
I need an example of a controller, which it upload a file. 
i.e. I want modify a image, so I have to delete and create a new.
I have seen a option 

```
 delete_on_update: true
```

But how does this work?

Thanks! : )

Koc:
Code is not tested yet, WIP.

1. use constants instead of strings
2. use one event instead of two
3. do not call recompute changeset because changesets not yet computed when `preFlush` triggers
4. fixes #89 

Looks like `AdapterInterface::recomputeChangeSet` is not called anymore.


EventListener/Doctrine/UploadListener.php
```diff
@@ -2,61 +2,43 @@
...
-     * @return array The array of events.
+     * {@inheritdoc}
      */
     public function getSubscribedEvents()
     {
-        return array(
-            'prePersist',
-            'preUpdate',
-        );
+        return array(Events::preFlush);
```
stof: this breaks for people using the ODM rather than the ORM as they won't have the constant defined

EventListener/Doctrine/UploadListener.php
```diff
@@ -2,61 +2,43 @@
...
     }
 
     /**
-     * @param EventArgs $event The event.
+     * @param PreFlushEventArgs $event
      */
-    public function prePersist(EventArgs $event)
+    public function preFlush(PreFlushEventArgs $event)
```
stof: this typehint will break as well for other Doctrine projects

EventListener/Doctrine/UploadListener.php
Koc: What you suggest: use strings or create different classes wich rely on different class contsants?

Koc:
@stof what can you say about this PR in general? And I have some problem with Propel integration. Not sure this approach will work correctly

EventListener/Doctrine/UploadListener.php
stof: use strings

EventListener/Doctrine/UploadListener.php
stof: typehint using only the classes from Doctrine Common (and add instanceof checks internally if they have not settled on a common API for the ``preflush`` event but this should also be reported to Doctrine as events have been more or less standardized in doctrine/common, but Doctrine\Common\EventArgs should extend ``\Doctrine\Common\Persistence\Event\ManagerEventArgs`` in future versions to be consistent with other events)

preflush is not properly standardized among Doctrine projects yet (check the case of the ODM fully then btw, as this event might cause you some issues)

EventListener/Doctrine/CleanListener.php
```diff
@@ -2,43 +2,47 @@
...
     /**
-     * @param EventArgs $event The event.
+     * @param PreFlushEventArgs $event
      */
-    public function preUpdate(EventArgs $event)
+    public function preFlush(PreFlushEventArgs $event)
     {
-        $object = $this->adapter->getObjectFromArgs($event);
+        $em = $event->getEntityManager();
+        $identityMap = $em->getUnitOfWork()->getIdentityMap();
```
stof: This is an internal Doctrine API AFAIK. Relying on it may not be the best idea

EventListener/Doctrine/InjectListener.php
```diff
@@ -2,42 +2,40 @@
 
 namespace Vich\UploaderBundle\EventListener\Doctrine;
 
-use Doctrine\Common\EventArgs;
+use Doctrine\ORM\Event\LifecycleEventArgs;
```
stof: you should use the doctrine/common class

garak:
@Koc can you solve conflict? And maybe adding some tests?

Koc:
@garak yes, will do it later after fiinish my other PRs. Also I should apply changes requested by @stof . This PR is much harder than other.

EventListener/Doctrine/CleanListener.php
Koc: How can I got changed entities in other way? `PreFlush` event not recieve entities

Koc:
branch restarted and rebased. I'm afraid that this PR breaks BC. But we can start working on 2.0 where BC-breaks are acceptable.

PR #623, opened 2016-09-04
Add annotations

Koc:
see https://github.com/symfony/symfony/pull/12299 for more information about event annotation


Koc:
@K-Phoen ready for review

PR #636, opened 2016-11-16
Avoid loading all metadata when there is no propel mapping

stof:
The Compiler pass registers listeners when the mapping used by the class is a Propel mapping.
But if there is no such mapping, there is no need to load the metadata at all.

The loading of metadata relies on an unsupported usage of Symfony (getting a service instance during the container building phase), and that it breaks in Symfony 3.2 (one of the dependencies of the ``vich_uploader.metadata_reader`` service depends on a cache pool on 3.2, which is not yet ready to be instantiated at the time your compiler pass runs).
Thanks to this change, the broken behavior only happens for people using the Propel integration (next plan is to look for a way to achieve the necessary behavior without instantiating the service here, to solve it for them). anyone not using the Propel 1 integration would have a working project again, as the compiler pass would return before reaching the broken case.

### DIFF
--- a/DependencyInjection/Compiler/RegisterPropelModelsPass.php
+++ b/DependencyInjection/Compiler/RegisterPropelModelsPass.php
@@ -23,12 +23,25 @@ class RegisterPropelModelsPass implements CompilerPassInterface
             return;
         }
 
+        $mappings = $container->getParameter('vich_uploader.mappings');
+        $hasPropelMapping = false;
+
+        foreach ($mappings as $mapping) {
+            if ($mapping['db_driver'] === 'propel') {
+                $hasPropelMapping = true;
+                break;
+            }
+        }
+
+        if (!$hasPropelMapping) {
+            return;
+        }
+
         $serviceTypes = array(
             'inject', 'clean', 'remove', 'upload',
         );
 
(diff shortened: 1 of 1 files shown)


Koc:
We can force people describe required mapped classes in config. For example say that current version support Symfony < 3.2 and if you want get support of newer version - please define your propel classes inside config.

stof:
this is one of the ideas I had. I will experiment with it. I also need to check whether there is a way to subscribe to any Propel model instead of only specific ones (similar to what is done when using Doctrine), which would avoid this need.
but this is left for a subsequent PR. This PR focuses on avoiding doing lots of useless (broken) work for people not using Propel.

Koc:
I've do research some time ago and as I understand - willdurand/propel-eventdispatcher-bundle not allow this. But https://github.com/glorpen/GlorpenPropelBundle supports

stof:
yeah, and there is already a PR adding support for it, which does not suffer from the Symfony 3.2 bug (it includes a refactoring similar to the one I submitted here btw)

IndraGunawan:
i use symfony 2.7 and found this issue too.

https://github.com/symfony/symfony/issues/20234

it would be great if this PR as patch current version :)

PR #657, opened 2017-01-03
Use asset function in form template

matthieumota:
Be sure link to file is always good if app is in root path or subfolder like preco http://symfony.com/doc/current/templating.html#linking-to-assets

### DIFF
--- a/Resources/views/Form/fields.html.twig
+++ b/Resources/views/Form/fields.html.twig
@@ -12,7 +12,7 @@
         {% endif %}
 
         {% if download_uri is defined and download_uri %}
-        <a href="{{ download_uri }}">{{ 'download'|trans({}, 'VichUploaderBundle') }}</a>
+        <a href="{{ asset(download_uri) }}">{{ 'download'|trans({}, 'VichUploaderBundle') }}</a>
         {% endif %}
     </div>
 {% endspaceless %}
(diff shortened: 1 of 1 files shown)


Koc:
Not sure about it. Download url can be generated by routing

garak:
I'm not sure to understand what "preco" is.
Anyway, what is the path that you can't currently get working using `uri_prefix` option?

MatthieuMota:
Sorry for preco, I mean "recommend". For me, uri_prefix must be relative to base url. Example:
My website production is http://www.startup.com/ and local env is http://localhost/customer/startup/ or http://startup.dev (depends on developer workflow :)). uri_prefix is /uploads/users

In form template from this bundle, download_url return : 
production : http://www.startup.com/uploads/users GOOD
dev 1 :  http://startup.dev/uploads/users GOOD
dev 2 :  http://localhost/uploads/users NOT GOOD but with asset function from twig it transforms to http://localhost/customer/startup/uploads/users

Ok I can redefine download_url in render method from controller but I'm very lazy :)

garak:
It looks like this PR makes sense. As @Koc mentioned, there are some use cases where this is not appropriate, but such cases could be covered by a template override.
In the end, we must choose what is the more general case and cover it with the default template, leaving special cases to custom templates.
My opinion is to favor `asset()`.

Koc:
Oh, now I understand the problem. +1 for `asset` by default if there is easy way to not apply it for some cases (introduce additional option to disable it?)

garak:
We could add another option to `VichFileType`, something like `asset_helper`, to control that.
If we do, the question is again: should we default such option to `false` or `true`? In this case, I'm more inclined for `false`, since is totally BC.

Koc:
lets make it `false` for BC now and change to `true` on 2.0

garak:
@MatthieuMota do you mind to add such configuration to this PR?

MatthieuMota:
@garak Ok I try to see this as soon I've time :)

MatthieuMota:
@garak Need to use 'assets.packages' service from Sf 3 but BC issue. Because it's 'templating.helper.assets' service on Sf2. I don't want to inject the complete service container in form to check if I use one of the two services according to symfony version. I wait your advice for that :)

garak:
I just checked on a project of mine, Symfony 2.8.15, assets.packages is there.

Koc:
for what you need this service directly? you can pass `asset_helper` to template and use it for calling or not calling `asset` function

MatthieuMota:
The service permit to generate the asset url directly without write any logic in the view. But I can pass it to the view if it's ok for you ?

garak:
@MatthieuMota in the meantime we added some indentations, could you please fix conflicts?

PR #660, opened 2017-01-05
No need translator injection in form type

MatthieuMota:
Translator is injected in form type service. I think it's not necessary because label form is translate and we can set custom translation domains.
More, It causes bug with symfony profiler who detects the string translate as an ID from domain "messages" in AppBundle.


garak:
Thank you!

PR #677, opened 2017-03-03
Create VichUploaderBundle.cs.yml

neonovic:
Czech language translations.


garak:
Thank you!

garak:
I just tagged version `1.5.1`, so you can use the new translation immediately

PR #688, opened 2017-03-24
Remove proxy-manager dependency

Koc:
Haven't see any usage of the proxy-manager in this bundle. Let's try remove and build on Travis.


Koc:
Travis build pass ok. Does we really need this dependency @K-Phoen, @garak ?

PR #690, opened 2017-03-27
Create embeddable entity for storing common file information

Koc:
Another one small step to reduce boilerplate when using the bundle.

TODO:

- [x] update documentation

### DIFF
--- a/Entity/File.php
+++ b/Entity/File.php
@@ -1,0 +1,95 @@
+<?php
+
+namespace Vich\UploaderBundle\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Embeddable
+ */
+class File
+{
+    /**
+     * @ORM\Column(name="name", nullable=true)
+     */
+    protected $name;
+
+    /**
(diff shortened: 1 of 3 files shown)


Koc:
@garak ready

Entity/File.php
```diff
@@ -0,0 +1,71 @@
...
+     * @ORM\Column(name="mime_type", nullable=true)
+     */
+    protected $mimeType;
+
+    /**
+     * @ORM\Column(name="size", type="integer", nullable=true)
+     */
+    protected $size;
+
+    public function getName()
```
garak: What about some phpdoc here? And on other methods, too

Koc: I can add phpdocs with param/return types, but I haven't see any profit from method descriptions like `Sets the original file name`. It clear from method name what it do :)

Koc: @garak done

Koc:
@garak yes, I'l do it later when other codebase will be moved to scalar type hints. Currently we are using this bundle on php 5.5.9 🤣

PR #691, opened 2017-03-27
Allow pass download_uri value into form types

Koc:
Useful for serving protected files via controller.


garak:
What about docs?

Koc:
@garak ready to review and merge. I've added note about new option into docs.

PR #700, opened 2017-04-16
Rework form options, download handler options

Koc:
Notable changes:
- form theme is automatically registered now - no config changes required (DX, also needed for https://github.com/symfony/recipes-contrib/pull/22 )
- `VichFileType`: `download_link` deprecated with `download_uri`, which supports callable
- `VichFileType`: `download_label` now supports `true` (use original file name), `callable` and instances of `PropertyPath`
- `VichFileType`: unified `translation_domain` option usage
- `VichImageType`: added option `image_uri`
- `VichImageType`: added optional integration with LiipImagineBundle
- download handler now supports `true` (use original file name) as served file name
- fixed phpdocs for `StorageInterface`, `UploaderHelper`, `DownloadHandler`
- reorganized test suite: validate docs separately from unit tests

### DIFF
--- a/Form/Type/VichFileType.php
+++ b/Form/Type/VichFileType.php
@@ -9,11 +9,21 @@ use Symfony\Component\Form\FormEvent;
 use Symfony\Component\Form\FormEvents;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\Form\FormView;
+use Symfony\Component\OptionsResolver\Options;
 use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\Component\PropertyAccess\PropertyAccess;
+use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
+use Symfony\Component\PropertyAccess\PropertyPath;
 use Vich\UploaderBundle\Form\DataTransformer\FileTransformer;
 use Vich\UploaderBundle\Handler\UploadHandler;
+use Vich\UploaderBundle\Mapping\PropertyMappingFactory;
 use Vich\UploaderBundle\Storage\StorageInterface;
 
+/**
+ * @author Kévin Gomez <contact@kevingomez.fr>
+ * @author Konstantin Myakshin <koc-dp@yandex.ru>
+ * @author Massimiliano Arione <max.arione@gmail.com>
+ */
 class VichFileType extends AbstractType
 {
     /**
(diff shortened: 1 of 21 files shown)


Resources/doc/form/vich_file_type.md
```diff
@@ -1,4 +1,4 @@
-VichFileType
+VichFileType Field
 ============
```
garak: You should adapt length here to match title length

Koc:
@garak are minior bc-breaks in form types template acceptable for 1.6.0?

Koc:
Also I can add integration with LiipImagineBundle for `ImageType`, wdyt?

garak:
Please alaborate about BC breaks and LiipImagineBundle

Koc:
BC-break - I've removed `show_download_link` view variable from `fields.twig`.

Integration: we can add option like `imagine_pattern` for getting images resized via LiipImagineBundle for `VichImageType` form type.

garak:
We should keep the variable, even if unused, to avoid breaking bad (I mean, causing errors). If the only break is a change if behaviour, it could be feasible.
The integration sounds good to me, as long as it's optional

Resources/doc/form/vich_image_type.md
```diff
@@ -19,27 +19,137 @@ class Form extends AbstractType
...
+**type**: `bool`, `string`, `callable` **default**: `true`
+
+If set to `true`, download uri will automatically resolved using storage.
+
+Can be string
+
+```php
+use Vich\UploaderBundle\Form\Type\VichFileType;
+
+$builder->add('genericFile', VichFileType::class, [
```
Koc: copy-paste error, need to be `VichImageType`

Form/Type/VichFileType.php
```diff
@@ -43,14 +49,32 @@ public function configureOptions(OptionsResolver $resolver)
...
 
         $resolver->setAllowedTypes('allow_delete', 'bool');
-        $resolver->setAllowedTypes('download_link', 'bool');
+        $resolver->setAllowedTypes('download_link', ['null', 'bool']);
+        $resolver->setAllowedTypes('download_uri', ['bool', 'string', 'callable']);
+        $resolver->setAllowedTypes('download_label', ['bool', 'string', 'callable', PropertyPath::class]);
         $resolver->setAllowedTypes('error_bubbling', 'bool');
+
+        $downloadUriNormalizer = function (Options $options, $downloadUri) {
+            if (isset($options['download_link'])) {
```
Koc: need comparation with null, `isset` not works as expected

garak: What if key is not set? Maybe an `empty` could be better?

Koc: It always set because of `setDefault` call

Koc:
@garak tests pass, all new cases are covered with tests, docs updated, please review and merge.

Form/Type/VichFileType.php
```diff
@@ -9,31 +9,37 @@
...
 use Vich\UploaderBundle\Storage\StorageInterface;
 
+/**
+ * @author Kévin Gomez <contact@kevingomez.fr>
+ * @author Konstantin Myakshin <koc-dp@yandex.ru>
+ * @author Massimiliano Arione <max.arione@gmail.com>
+ */
 class VichFileType extends AbstractType
 {
-    /**
```
garak: Is there a reason for removing phpdoc blocks here (and in the following properties/methods)?

Koc: This blocks are duplication of real typehints in constructor so there is no any benefit from they IMHO.

garak: I'm not sure that every IDE is smart enough to pass the type after assignment. Anyway, phpdoc block could be used for other than autocompletion in IDEs (e.g. for doc auto-generation).
Anyway, it would be better to keep consistency between files (so, since other files do use phpdoc blocks, this file should use it as well)

Koc: restored

Issue #692, opened 2017-03-27
Setting delete_on_update to false does not work as expected


cristiangsp:
Hi,

In our use case, we don't want the Bundle to delete the file from our bucket on S3 when it's replaced by other one in a Doctrine Document (we use MongoDB). We expected that the delete_on_update configuration parameter set to false on that specific mapping would be enough to make it work that way. But what happens s that the files just get deleted. 

After digging a little bit into the Bundle structure it seems that the delete_on_update flag is only taken into account when registering the listeners associated to the database driven. The problem is that the CleanListener is triggered no matter what as it seems associated to the preUpload event that Doctrine trigger.

Here is our mapping configuration:

```yml
        profile_common_avatar:
            uri_prefix: https://s3.amazonaws.com/%aws.s3.bucket%/profiles/avatars
            upload_destination: profile_common_user_avatar_images
            namer: vich_uploader.namer_uniqid
            delete_on_update: false
```

To double check, we manually blocked any delete operation in one of our S3 Buckets for testing and the Bundle ends up throwing an exception related to the fact that it cannot delete the element even though it should not try as the delete_on_update is set to false. 

Here is the Stack Trace where it's clear how the CleanListener is being triggered:

```code
1. in vendor/knplabs/gaufrette/src/Gaufrette/Filesystem.php at line 170   -
            return true;
        }
        throw new \RuntimeException(sprintf('Could not remove the "%s" key.', $key));
    }
2. at Filesystem ->delete ('58d926daae49c.jpeg') 
in vendor/vich/uploader-bundle/Storage/GaufretteStorage.php at line 69   +
3. at GaufretteStorage ->doRemove (object(PropertyMapping), '', '58d926daae49c.jpeg') 
in vendor/vich/uploader-bundle/Storage/AbstractStorage.php at line 94   +
4. at AbstractStorage ->remove (object(CommonProfile), object(PropertyMapping)) 
in vendor/vich/uploader-bundle/Handler/UploadHandler.php at line 111   +
5. at UploadHandler ->remove (object(CommonProfile), 'photo') 
in vendor/vich/uploader-bundle/Handler/UploadHandler.php at line 102   +
6. at UploadHandler ->clean (object(CommonProfile), 'photo') 
in vendor/vich/uploader-bundle/EventListener/Doctrine/CleanListener.php at line 40   +
7. at CleanListener ->preUpdate (object(PreUpdateEventArgs)) 
[...]

PR #707, opened 2017-05-06
Added test for #692 and fix some method

Koc:
try cover #692 with test and implements support of the second argument for `getUploadableFields` (it called with 2 arguments in some places)


garak:
Do you want to squash yourself? Using the automatic squashing tool here on github, the message is usually messy

Koc:
@garak done

PR #711, opened 2017-05-17
Add doctrine/annotations dependency

Koc:
see
https://github.com/symfony/symfony/pull/20097
https://github.com/symfony/recipes-contrib/pull/22#issuecomment-301831276

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,7 @@
     },
     "require": {
         "php": "^7.0",
+        "doctrine/annotations": "^1.3",
         "symfony/dependency-injection": "^2.8|^3.1",
         "symfony/framework-bundle": "^2.8|^3.1",
         "symfony/finder": "^2.8|^3.1",


harmstyler:
Is this necessary as a required dependency? Due to the fact that Annotations are optional for this bundle the annotations library should be optional as well. I'd recommend a composer suggestion over a requirement:
```json
"suggest": {
    "doctrine/annotations": "To use Symfony annotations."
}
```

garak:
I agree.
Likely we'll need a line in "require-dev" also

Koc:
Doctrine is a first-class citizen. Adding this dependency as required helps us avoid many missunderstandings and issues in future.

garak:
I think it's just a matter of begin clear in docs.
Even if annotations are suggested in first place, they're not technically a requirement

Koc:
Some parts of DI should be rewrited. Currently annotations are required by code, see https://github.com/symfony/recipes-contrib/pull/22#issuecomment-301831276

garak:
If so, annotations are actually a requirement.

PR #714, opened 2017-05-17
Added symfony validation for non ascii characters

cristobal151:
I have a problem with the character º in a filename. Does not pass symfony validations to accept only ascii characters

Symfony make this validation for non ascii characters for filename. See the code [here](https://github.com/symfony/symfony/blob/3.3/src/Symfony/Component/HttpFoundation/ResponseHeaderBag.php#L287)

So, I added a replace with the same regex in the funcion Vich\UploaderBundle\Util\Transliterator::transliterate

### DIFF
--- a/Tests/Handler/DownloadHandlerTest.php
+++ b/Tests/Handler/DownloadHandlerTest.php
@@ -130,7 +130,7 @@ class DownloadHandlerTest extends TestCase
             ->with($this->object, 'file_field')
             ->will($this->returnValue('something not null'));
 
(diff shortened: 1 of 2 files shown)


garak:
We need a test for this change.
Please add the offending character to `testNonAsciiFilenameIsTransliterated` method in `DownloadHandlerTest`

cristobal151:
Done, character º added in test.

garak:
Thanks

cristobal151:
:)

PR #715, opened 2017-05-23
Fix transliterator refactoring

Koc:
Remove not needed inheritance and fix pathinfo for cyrilic filenames https://3v4l.org/pCkom

Waiting for Travis

### DIFF
--- a/Tests/Util/FilenameUtilsTest.php
+++ b/Tests/Util/FilenameUtilsTest.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Vich\UploaderBundle\Tests\Util;
+
+use PHPUnit\Framework\TestCase;
+use Vich\UploaderBundle\Util\FilenameUtils;
+
+class FilenameUtilsTest extends TestCase
+{
+    /**
+     * @dataProvider spitNameByExtensionProvider
+     */
+    public function testSpitNameByExtension($filename, $basename, $extension)
+    {
+        $this->assertSame([$basename, $extension], FilenameUtils::spitNameByExtension($filename));
+    }
+
+    public function spitNameByExtensionProvider()
+    {
+        return [
+            'simple filename with extension' => ['filename.extension', 'filename', 'extension'],
+            'cyrillic filename with extension  ' => ['Текстовый файл.txt', 'Текстовый файл', 'txt'],
+            'cyrillic filename with dot and extension' => ['Текстовый .файл.txt', 'Текстовый .файл', 'txt'],
+            'cyrillic filename without extension ends with dot' => ['Текстовый файл.', 'Текстовый файл', ''],
+            'cyrillic filename without extension' => ['Текстовый файл', 'Текстовый файл', ''],
+        ];
+    }
+}
--- a/Util/Transliterator.php
+++ b/Util/Transliterator.php
@@ -2,10 +2,17 @@
 
 namespace Vich\UploaderBundle\Util;
 
-use Behat\Transliterator\Transliterator as BaseTransliterator;
+use Behat\Transliterator\Transliterator as BehatTransliterator;
 
-class Transliterator extends BaseTransliterator
+class Transliterator
 {
+    /**
+     * This class should not be instantiated.
+     */
+    private function __construct()
+    {
+    }
+
     /**
      * Transliterate a string. If string represents a filename, extension is kept.
      *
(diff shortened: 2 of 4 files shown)


Koc:
@garak done

garak:
What is done? I don't see any of my comments addressed

Koc:
I mean that the PR is done :)

Koc:
@garak can you review/merge?

garak:
I already reviewed about a week ago

Koc:
so can we got merged?

garak:
Not yet: review is still pending, awaiting your replies

Koc:
I haven't seen any comments or request changes 😕

garak:
My bad, the UI of github confused me, I was sure to have submitted it before... now is there

Tests/Util/FilenameUtilsTest.php
```diff
@@ -0,0 +1,27 @@
...
+     */
+    public function spitNameByExtension($filename, $basename, $extension)
+    {
+        $this->assertSame([$basename, $extension], FilenameUtils::spitNameByExtension($filename));
+    }
+
+    public function spitNameByExtensionProvider()
+    {
+        return [
+            ['Текстовый файл.txt', 'Текстовый файл', 'txt'],
```
Koc: I've added test set in English. Also you can refer for test method signature

Util/Transliterator.php
```diff
@@ -2,9 +2,9 @@
 
 namespace Vich\UploaderBundle\Util;
 
-use Behat\Transliterator\Transliterator as BaseTransliterator;
+use Behat\Transliterator\Transliterator as BehatTransliterator;
 
-class Transliterator extends BaseTransliterator
+class Transliterator
```
Koc: I'm prefer Symfony's approach with private constructor
- https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Form/Util/FormUtil.php#L22
- https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Config/Util/XmlUtils.php#L27

Util/Transliterator.php
```diff
@@ -16,10 +16,10 @@ class Transliterator extends BaseTransliterator
...
     public static function transliterate($string, $separator = '-')
     {
-        $extension = pathinfo($string, PATHINFO_EXTENSION);
-        $filename = pathinfo($string, PATHINFO_FILENAME);
-        $transliterated = parent::transliterate($filename, $separator);
-        if (!empty($extension)) {
+        list($filename, $extension) = FilenameUtils::spitNameByExtension($string);
+
+        $transliterated = BehatTransliterator::transliterate($filename, $separator);
+        if ($extension) {
```
Koc: Added strict comparasion with empty string

Tests/Util/FilenameUtilsTest.php
garak: Still I can't easily understand what is done in each of the elements of this array, since key are (implicit) numeric. If you replace them with string keys, it can be easier to understand. For example, the first element could have a key like 'filename with extension', the second element 'cyrillic filename with whitespace' (just a wild guess), and so on

Koc: We have a lot of data providers and no one has something similar that you talking about

- https://github.com/dustin10/VichUploaderBundle/blob/dcb416194e54e75def9d96b65c232e4346499a42/Tests/Naming/PropertyNamerTest.php#L24
- https://github.com/dustin10/VichUploaderBundle/blob/dcb416194e54e75def9d96b65c232e4346499a42/Tests/Naming/OrignameNamerTest.php#L17

each set contains file name, basename and extension

garak: And that's a shame, because if a test is failing you don't get any clue about what is failing

Koc: Signature of method which uses this data provider perfectly explains all the things:
`testSpitNameByExtension($filename, $basename, $extension)`

garak: But it does not explain the cases you're providing to it. You should imagine that instead of that single test you get 5 tests (one for each row provided by the data provider). In such situation, the 5 tests would get a different name explaining the single case you're testing (and maybe what you expect). Since there is only one test, we need to move such help to the data provider, otherwise a possible failing test would tell not enough.
See example 2.6 on https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers

Koc: ah, now I understand what do you want, ok, done

garak:
Thank you!

Koc:
@garak can you release 1.6 please? I want start working on 2.0

Also some tags missed real releases with some changelog notes https://github.com/dustin10/VichUploaderBundle/releases

PR #724, opened 2017-06-13
Add options to not load form and templating services

akeeman:
This makes loading of form and templating services optional, like it's already optional to load twig stuff.
This helps minimize APIs, and also offers a ways to work around requiring `form.property_access`, that is a dependency of this bundle since 1.6.0.


garak:
LGTM, but I think we need to mention this in documentation.
Can you update your PR to include it? Thanks

akeeman:
I've updated the docs so it conforms to what's written about the twig enable/disable option, and made it a bit more clear. Also, I've added a dependency of twig on templating, as the twig extension service uses the templating service.

teohhanhui:
We can use `class_exists` and `interface_exists` to provide the correct defaults. That would work better for Symfony Flex.

PR #725, opened 2017-06-14
[POC] Make yaml and annotations are optional

Koc:
Just proof of concept for now. Continuation of discussion https://github.com/dustin10/VichUploaderBundle/pull/711#issuecomment-307930721


Koc:
//cc @harmstyler

PR #727, opened 2017-06-14
Test on Symfony 3.3 and integrate phpunit-bridge

Koc:
two questions before merge
1. do we need leave 3.2 also or even 3.1 because of composer reqs?
2. do we need add test env without full-stack and only for framework-bundle (or `--prefer-lowest`)?


Tests/Vich/UploaderBundle/Tests/Form/Type/VichFileTypeTest.php
```diff
@@ -21,8 +21,6 @@ class VichFileTypeTest extends TestCase
 
     /**
      * @dataProvider configureOptionsBCDataProvider
-     * @group legacy
-     * @expectedDeprecation The "download_link" option is deprecated since version 1.6 and will be removed in 2.0. You should use "download_uri" instead.
```
Koc: removed just for testing that build fails without this annotations

Koc:
@garak ready, please review

Tests/PHPUnitTestCase.php
```diff
@@ -0,0 +1,5 @@
+<?php
```
garak: What's the purpose of this file?

composer.json
```diff
@@ -42,7 +42,9 @@
         "symfony/browser-kit": "^2.8|^3.1",
         "symfony/css-selector": "^2.8|^3.1",
         "symfony/dom-crawler": "^2.8|^3.1",
+        "symfony/validator": "^2.8|^3.1",
```
garak: unsorted

Tests/PHPUnitTestCase.php
Koc: It required for test pass on `2.8.0` when composer run with `--prefer-lowest`. Symfony added support of the namespaced PHPUnit from 2.8.18

garak: So it's better to raise requirements for symfony to `^2.8.18`. There's no value in supporting older versions

Koc: If we can do it - of course, yes

garak: go for it

garak:
There's no need to drop Symfony 3.1 and 3.2 for now, even if we require the phpunit bridge for 3.3 (since it's compatible with previous Symfony versions).
Anyway, we need to investigate on the failure for the lowest requirements and see if possibly increase some dependencies' minimum versions

Koc:
@garak tests are green finally now

Issue #721, opened 2017-06-12
Yaml Config class name incorrect


harmstyler:
As of the latest release, 1.6.0, the Yaml upload mappings no longer work. It appears that the generated class name built by the bundle is no longer correct.

For example, running the following command in `bin/console vich:mapping:list-classes` returns different results depending tag I'm on.

1.5.3: `AppBundle\Entity\Avatar`
1.6.0: `AppBundle\Avatar`

I tried to track down the breaking change but the 1.6.0 release was rather large, and not entirely captured in the release docs. I believe, however, the offending change happened in this commit: e6706b22b98e3d52f8267d7b2e4095adc633dbac, though I'm not positive..

Koc:
Referenced commit fixes another bug: #704. Can you create fork of symfony-standard edition with reproducible env for this bug?

Koc:
Also this changes covered with new tests https://github.com/dustin10/VichUploaderBundle/commit/c40cdae8371044ebe010b2c25ad3f38525e1ee70 . Could you provide own fixtures and tests?

flochntrl:
I have the same problem with Symfony 3.2.9

Koc:
again, please, create small piece of reproducible code for this bug by forking symfony-standard

Koc:
@harmstyler, @flochntrl can you check #728, does it solves your problem?

flochntrl:
@Koc LGTM ! Thanks :+1:

shairyar:
I am running into the same issue, I have provided more details [here](https://stackoverflow.com/questions/44653118/symfony-vichuploaderbundle-not-working-after-composer-update) incase it helps. This bundle seems to be a dependency of another bundle i am using `presta/image-bundle` its not listed in my composer.json file so how do i downgrade it to 1.5.3

PR #728, opened 2017-06-14
Document BC-break, introduced in 1.6.0

Koc:
Possible fix for #721


harmstyler:
Though this might fix the initial problem with yaml/xml mappings that was introduced in 1.6.0, I'm not sure this is the best permanent solution. Changing the file names to match directory structure seems to make the existing yaml config redundant. We are already storing the full Class name as the config name in the files in question. Thoughts?

Koc:
When we need get mapping for concrete class name we need to know what file to include. So files should be named according some naming convention.

Koc:
@garak what your opinion on this situation? There is at least one comment which confirms that described in this PR migration plan helps https://github.com/dustin10/VichUploaderBundle/issues/721#issuecomment-308967040

PR #750, opened 2017-07-11
Added info about Gaufrette with Doctrine adapter

kucharovic:
- Added information about Gaufrette with Doctrine adapter dependency on `ocramius/proxy-manager`. 
- Added sugggestion to `composer.json`


garak:
Thank you!

rogamoore:
This should be in master in bold letters. I spent hours trying to debug this.

garak:
I'm not sure how/when/why suggestion was lost from master. Anyway, I just re-added it.

PR #755, opened 2017-07-24
Add "delete_callback" option to form types.

Koc:
Do we need cover this with test?


garak:
Everything needs to be covered with tests

kimwue:
Would like this feature.

Koc:
@kimwue nice to see anybody else who has similar problems like me :). Will try finish this PR soon.

garak:
I'm about to close this, since it's a bit old.
Also, can't you just use a listener to reach the same goal?

Koc:
This PR is not relevant anymore

PR #765, opened 2017-09-13
Document naming convention for .yml config files

benr77:
Added some additional information on how the YAML entity files should be names, as it must correspond with the namespace_prefix setting in order for the file to be found correctly.


garak:
Thanks!

PR #792, opened 2017-11-12
Base64Namer

Tejes:
I wanted a namer in my application which creates shorter IDs similar to YouTube video IDs. I hope it can be useful to others.
The implementation is based on HashNamer, but since this bundle is targeting PHP7, I used the random_int() function instead. The name is URL-safe and has configurable length.


garak:
Thank you @Tejes

PR #806, opened 2017-12-25
Increase requirements

garak:


Tests/Mapping/PropertyMappingTest.php
```diff
@@ -33,52 +33,10 @@ public function testConfiguredMappingAccess()
         $this->assertEquals('fileName', $prop->getFileNamePropertyName());
     }
 
-    /**
```
Koc: why removed?

garak: Because enforcing parameter type is making tests failing. Of course, if you know a better way, I'm open to suggestions

PR #812, opened 2018-01-11
Uses FileExtensionTrait in all files namers.

phpeek:
It appears that files namers are currently determining file extension in different ways.

This PR unifies it by using `Polyfill\FileExtensionTrait` in all file namers.

### DIFF
--- a/Naming/Base64Namer.php
+++ b/Naming/Base64Namer.php
@@ -12,6 +12,8 @@ use Vich\UploaderBundle\Mapping\PropertyMapping;
  */
 class Base64Namer implements NamerInterface, ConfigurableInterface
 {
+    use Polyfill\FileExtensionTrait;
+
     protected const ALPHABET = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ-_';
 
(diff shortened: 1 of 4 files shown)


garak:
What about OrignameNamer.php?

phpeek:
@garak `OrignameNamer` has never used `FileExtensionTrait` since it's basing its behavior solely on `$file->getClientOriginalName()`.

garak:
That sounds weird to me.
So, using `OrignameNamer` I'll get the original file name without the extension?

phpeek:
I know. It was a bit weird to me too, when I first saw it, but decided to leave it as is.

You'll get an extension though it's gonna be original extension derived from the original file name, not guessed by `->guessExtension()`.

Behavior of the trait is as follows -

1. Try to extract extension for the original file name by using - `pathinfo($originalName, PATHINFO_EXTENSION)`. If that's not possible due to original filename not having an extension in its name, fallback to 2.

2. Extract extension by guessing it from file's `mimeType`.

`OrignameNamer` never did 2. step, so if original file name don't have extension in its name, then renamed file name won't have an extension in its name too.

I know it's a bit complicated and hard to describe. I'll work on utilizing it all, but that's for another PR.

garak:
Thank you!

PR #815, opened 2018-01-15
Fixes typo in method's name and uses pathinfo to extract ext and name.

phpeek:
There's a typo in `spitNameByExtension` method.

Previously this method was using combination of `strpos` and `substr` to extract filename and extension. I've switched it to much simpler `pathinfo` and `??` operator approach.


phpeek:
Tests are failing because of `pathinfo()` being `locale` aware. Closing for now.

Issue #810, opened 2018-01-10
Add aliases for services


garak:
I think it could be nice if we add aliases that map services class names to services names.
Example: we currently expose a service named `vich_uploader.download_handler` that points to `Vich\UploaderBundle\Handler\DownloadHandler` (and it's pubilc)
We could add an alias like this:

```xml
 <service id="Vich\UploaderBundle\Handler\DownloadHandler" alias="vich_uploader.download_handler" public="false"/>
```

in this way, users are not forced to declare their own aliases to use such services in controllers.
Furthermore, we could deprecate public services, so in next major release we'll be able to remove then (and replace them with the aliases)

What do you think?

garak:
Ping @Koc

Koc:
+1 for download handler. And maybe create interface for it.

Storage also requires in some cases.

Koc:
Can we reopen it until https://github.com/dustin10/VichUploaderBundle/commit/aad4ccceb5bc6d832f27a025334334b159073f80#r27021884 comment not resolved? I'm afraid that we forgot about it.

garak:
It's already solved, see https://github.com/dustin10/VichUploaderBundle/blob/master/Resources/config/storage.xml#L11

PR #818, opened 2018-01-16
Add aliases to be able to extend VichImageType

juntereiner:
I extended the VichImageType class and I had to put aliases on 3 services to be able to use the autowiring.

Maybe we could also add alias for PropertyAccessorInterface and CacheManager too.

### DIFF
--- a/Resources/config/factory.xml
+++ b/Resources/config/factory.xml
@@ -10,5 +10,7 @@
             <argument>%vich_uploader.mappings%</argument>
             <argument>%vich_uploader.default_filename_attribute_suffix%</argument>
         </service>
+
+        <service id="Vich\UploaderBundle\Mapping\PropertyMappingFactory" alias="vich_uploader.property_mapping_factory" public="false"/>
     </services>
 </container>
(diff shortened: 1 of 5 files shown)


garak:
See #810

juntereiner:
Ok so you want all alias to be private and put deprecate notice on all public service?

garak:
Exactly

juntereiner:
I try to put deprecation notice on public service but It triggers even if you use it with the alias... which is logical when we think about it.

garak:
We need to investigate if there's a way to make deprecations dynamic

juntereiner:
As said [here](https://symfony.com/blog/new-in-symfony-2-8-deprecated-service-definitions), this feature was not implemented for alias.

What we could do is to create new service definitions with the FQCN and private, and put deprecated notice on older service. But I don't think it's a good idea,  we will double the number of services and lots of deprecation notices will trigger before new release version with BCs.

garak:
I see.
I found a [PR](https://github.com/symfony/symfony/pull/24707) on deprecating aliases, but it's stuck (and likely will be merged on 4.1 only). So we can't add deprecations in services for now, we'll need to state necessary changes in UPGRADE.

PR #820, opened 2018-01-17
Fix test environment for the new environment requirement

juntereiner:
Actually, you cannot do a composer install nor run the tests because of 2 things:

1 - php7.1 is now the minimum and doctrine/mongodb-odm needs ext-mongo which doesn't exist for php7.1 so you need to do this: http://docs.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/introduction.html#using-php-7

2 - you can't run the script runTests.sh because it tries to test SF 2.8 and 3.3 which is not possible anymore with the new requirements

Edit: travis failing because ext-mongodb is missing (which should be used with php7.1)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,9 +28,10 @@ matrix:
         - php: nightly
 
 before_install:
+    - if [ -z "$(pecl list | grep -w mongodb)" ]; then pecl install mongodb-1.3.4; fi
+    - echo "extension = mongodb.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
     - composer self-update
     - phpenv config-rm xdebug.ini || true
-    - composer config platform.ext-mongo 1.6.14
     - if [ "$SYMFONY_VERSION" != "" ]; then composer require --no-update symfony/symfony:${SYMFONY_VERSION}; fi
     - if [ "$WITH_LIIP_IMAGINE" = true ] ; then composer require --no-update liip/imagine-bundle:"^1.7|^2.0@dev"; fi;
     - if [ "$VALIDATE_DOCS" = true ]; then composer require --dev --no-update kphoen/rusty dev-master; fi
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,7 @@
     },
     "require-dev": {
         "ext-sqlite3": "*",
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -1,4 +1,5 @@
 .php_cs.cache
 phpunit.xml
 composer.lock
+composer.phar
```
garak: Please remove this

juntereiner: Done

garak:
```
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.
  Problem 1
    - alcaeus/mongo-php-adapter 1.1.3 requires ext-mongodb ^1.2.0 -> the requested PHP extension mongodb is missing from your system.
    - alcaeus/mongo-php-adapter 1.1.2 requires ext-mongodb ^1.2.0 -> the requested PHP extension mongodb is missing from your system.
    - alcaeus/mongo-php-adapter 1.1.1 requires ext-mongodb ^1.2.0 -> the requested PHP extension mongodb is missing from your system.
    - alcaeus/mongo-php-adapter 1.1.0 requires ext-mongodb ^1.2.0 -> the requested PHP extension mongodb is missing from your system.
    - Installation request for alcaeus/mongo-php-adapter ^1.1 -> satisfiable by alcaeus/mongo-php-adapter[1.1.0, 1.1.1, 1.1.2, 1.1.3].
  To enable extensions, verify that they are enabled in your .ini files:
    - /home/travis/.phpenv/versions/7.1.11/etc/php.ini
    - /home/travis/.phpenv/versions/7.1.11/etc/conf.d/travis.ini
  You can also run `php --ini` inside terminal to see which files are used by PHP in CLI mode.
```

juntereiner:
That's because we need the php extension ext-mongodb with php7.1, so you need to add that in your travis.

garak:
Can you fix `.travis.yml`?

garak:
It looks like we need to make mongodb installation dependent on PHP version

juntereiner:
Travis is happy now but I think we should put composer.phar in gitignore because when we run runTests.sh, It downloads a composer.phar but never removes it.

We should either put composer.phar in gitignore or remove it at the end of the script.

garak:
Thanks!

PR #828, opened 2018-01-26
Remove aliasing as it done in extension

Koc:
Sorry, I've forgot to remove it when moved to extension


garak:
What about other storages aliases?

Koc:
Interesting question. Maybe they should remains because they are useful inside config 

```yaml
vich_uploader:
    storage: '@Vich\UploaderBundle\Storage\FlysystemStorage'
```

Same question for form types. Does any other service can accept form type as argument? When it could be useful? I haven't see any use case.

Koc:
@garak so what's your verdict about storages and form types aliases?

garak:
I think we should keep as much aliases as we can.
The idea is that in the next major version we'll try to drop "normal" service names and use class names instead, to fully support autowiring

PR #831, opened 2018-01-29
Fix `imagine_pattern` option behavior in VichImageType

ossinkine:
`Liip\ImagineBundle\Imagine\Cache\CacheManager::getBrowserPath` should receive a relative path in the first argument but not an URI.
Moreover this works if `LiipImagineBundle` and `VichUploaderBundle` properly configured, config options `vich_uploader.mappings.default.upload_destination` and `liip_imagine.loaders.default.filesystem.data_root` must point to the same directory.
Current version might only work if URI that returned by the Storage equals to relative path for Imagine.

### DIFF
--- a/Form/Type/VichImageType.php
+++ b/Form/Type/VichImageType.php
@@ -61,9 +61,9 @@ class VichImageType extends VichFileType
                     throw new \RuntimeException('LiipImagineBundle must be installed and configured for using "imagine_pattern" option.');
                 }
 
-                $uri = $this->storage->resolveUri($object, $form->getName());
-                if ($uri) {
-                    $view->vars['image_uri'] = $this->cacheManager->getBrowserPath($uri, $options['imagine_pattern']);
+                $path = $this->storage->resolvePath($object, $form->getName(), null, true);
+                if (null !== $path) {
+                    $view->vars['image_uri'] = $this->cacheManager->getBrowserPath($path, $options['imagine_pattern']);
                 }
             } else {
                 $view->vars['image_uri'] = $this->resolveUriOption($options['image_uri'], $object, $form);
--- a/Tests/Form/Type/VichImageTypeTest.php
+++ b/Tests/Form/Type/VichImageTypeTest.php
@@ -146,9 +146,9 @@ class VichImageTypeTest extends VichFileTypeTest
         $storage = $this->createMock(StorageInterface::class);
         $storage
             ->expects($this->any())
-            ->method('resolveUri')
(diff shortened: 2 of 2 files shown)


Form/Type/VichImageType.php
```diff
@@ -61,9 +61,9 @@ public function buildView(FormView $view, FormInterface $form, array $options):
                     throw new \RuntimeException('LiipImagineBundle must be installed and configured for using "imagine_pattern" option.');
                 }
 
-                $uri = $this->storage->resolveUri($object, $form->getName());
-                if ($uri) {
-                    $view->vars['image_uri'] = $this->cacheManager->getBrowserPath($uri, $options['imagine_pattern']);
+                $path = $this->storage->resolvePath($object, $form->getName(), null, true);
+                if ($path) {
```
garak: Can we avoid implicit casting here? `null !== $path` would be better

ossinkine: Done

garak:
Thank you!

ossinkine:
@garak please prepare a new release with this fix

garak:
@ossinkine what do you think?

ossinkine:
@cadavre since `getBrowserPath` receives a relative path the `resolvePath` must also return the relative path (filename in a particular case). `uri_prefix` should not affect this in any way.
If you claim that it works with empty prefix, but does not work with not empty prefix, then please create a minimal example I be able to reproduce it.

leblanc-simon:
@ossinkine I have the same problem as @cadavre 
Here, you can have an example : https://gist.github.com/leblanc-simon/d610160d70a81cf830175906431a5a74

This script : 
* install last Symfony LTS
* install and configure VichUploader and LiipImagine
* create the Entity and the Controller
* you must create a database before to launch this script

You can add an picture in the form and in the next page, you have a 404 error because the source file is not found.

If I revert the commit, all is OK.

cadavre:
Thank you @leblanc-simon for such a complete example.

ossinkine:
I'm sorry for delay, I was busy. @leblanc-simon Thank you for such detailed example.
Since you set `vich_uploader.mappings.image.upload_destination` parameter in `config.yml` you should also set this path for `liip_imagine`:
```yaml
liip_imagine:
    loaders:
        default:
            filesystem:
                data_root: '%kernel.project_dir%/web/uploads/images'
```

leblanc-simon:
Thank you very much @ossinkine !
With this configuration, all is OK

PR #876, opened 2018-06-12
Make docs more clear.

marcharding:
Make it clear to use the VichFileType  and not the native Symfony FileType.

### DIFF
--- a/Resources/doc/usage.md
+++ b/Resources/doc/usage.md
@@ -321,7 +321,7 @@ All options are listed below:
 
 ## That was it!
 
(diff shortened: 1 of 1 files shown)


garak:
Why not `VichImageType`?

marcharding:
Yes, even better 😉 

The docs are a bit confusing in this regard, i still had the native symfony file type in my form and wondered why it did not work. I suppose that could happen to others as well..

garak:
Can you update your PR?

garak:
Thank you @marcharding !

PR #880, opened 2018-06-19
Implement a queue for removing files during postFlush event

### LINKED ISSUE #756: File can be removed physically from storage without db update
I've found interesting edge case: create form with vich file type (allow delete = true), then check remove file and submit invalid form. Post submit listener removes file from disk and clean file information from entity. But usually nobody calls `flush` for invalid forms. So we got unsynchronized state: file was removed but database not updated.

I see 2 different approaches to fixing this issue:
a. change priority of our removing listener and check for `isValid` first (easy but not cover all cases)
b. add dynamic preflush listener to entity manager wich will remove file

wuestkamp:
This fixes #756 by implementing a queue in which we store entities + fieldnames. On postFlush we call the already existing "remove" method on all queue entries.

Updated this PR with unit tests.

I created a small test project to test this error:
https://github.com/wuestkamp/VichUploaderBundleTest756/blob/master/README.md

Kim


wuestkamp:
Propel Flush Listener is still missing, but working for doctrine. According to [this](https://github.com/dustin10/VichUploaderBundle/blob/master/Resources/doc/propel.md) Propel2 is not supported?

EventListener/Doctrine/FlushListener.php
```diff
@@ -0,0 +1,46 @@
...
+     *
+     * @param PostFlushEventArgs $event The event
+     */
+    public function postFlush(PostFlushEventArgs $event): void
+    {
+        $em = $event->getEntityManager();
+        $entities = $this->handler->removeFilesInQueue();
+
+        if (count($entities) > 0) {
+            foreach ($entities as $entity) {
```
garak: Are we really needing to persist here? Entities should be already in the entity manager

Handler/RemoveHandler.php
```diff
@@ -0,0 +1,125 @@
...
+use Vich\UploaderBundle\Storage\StorageInterface;
+use SplObjectStorage;
+
+/**
+ * Remove handler.
+ *
+ * @author Kévin Gomez <contact@kevingomez.fr>
+ * @author Kim Wüstkamp <kim@wuestkamp.com>
+ */
+class RemoveHandler extends AbstractHandler
```
garak: This class is extending from abstract, so it should be final.

Tests/Handler/RemoveHandlerTest.php
```diff
@@ -0,0 +1,223 @@
...
+        $this->handler->addToQueue($object2, self::FILE_FIELD_TWO);
+
+        $objects = $this->handler->removeFilesInQueue();
+
+        $this->assertEquals(count($objects), 2);
+        $this->assertTrue($objects[0] === $object1);
+        $this->assertTrue($objects[1] === $object2);
+    }
+
+    protected function getStorageMock()
```
garak: I guess that this test class is not meant to be extended. Protected methods should be private instead

EventListener/Doctrine/FlushListener.php
wuestkamp: You're correct, no persist needed.

Tests/Handler/RemoveHandlerTest.php
```diff
@@ -0,0 +1,223 @@
...
+    /**
+     * @var RemoveHandler
+     */
+    protected $handler;
+
+    const FILE_FIELD_ONE = 'image';
+
+    const FILE_FIELD_TWO = 'attachment';
+
+    private function setUp(): void
```
garak: This can't be private, since parent is protected

garak:
Rebase needed

garak:
This became really old, we need a rebase or this PR will be closed without merge

wuestkamp:
This MR was never merged when it was ready. It fixes a bug (https://github.com/dustin10/VichUploaderBundle/issues/756) which results in files being deleted even though the user doesn't want it. This should definitely be prioritised in my opinion. Sadly I won't have time to work again on this for now.

PR #888, opened 2018-07-09
Check if parent is null

nataliaAznar:
When using NelmioApiDocBundle, the forms parent is null.
For this reason, it is necessary tocheck if it is null before getting its data


garak:
Thank you @nataliaAznar

davidpeoples:
I hope this makes it into a tagged branch soon. I ran into this very problem yesterday.

garak:
@davidpeoples you can use `^1.8@dev` meanwhile

PR #892, opened 2018-07-26
Callback namers

Padam87:
Added callback namers. Objects using these namers must implement an interface, and return a file / directory name.

This would eliminate a huge percent of cases in which I had to create a namer.

Let me know if you are willing to merge this, and I will finish up docs / etc.

### DIFF
--- a/Naming/CallableDirectoryNameProviderInterface.php
+++ b/Naming/CallableDirectoryNameProviderInterface.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace Vich\UploaderBundle\Naming;
+
+interface CallableDirectoryNameProviderInterface
+{
+    public function getUploadedDirectoryName(): string;
+}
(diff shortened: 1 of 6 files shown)


garak:
This looks interesting. Could you please elaborate a bit on the use case?

Padam87:
I work in the printing industry, and have to deal with a lot of different files. What I noticed is that most of the time I needed to customize a namer is because I wanted to place the file in a directory according to a related entity. For example:

```php
class OrderProductFile extends \Vich\UploaderBundle\Entity\File implements CallableDirectoryNameProviderInterface
{
    // ...

    public function getUploadedDirectoryName(): string
    {
        return implode(
            DIRECTORY_SEPARATOR,
            [
                $this->design->getOrderProduct()->getOrder()->getId(),
                $this->design->getOrderProduct()->getId(),
                $this->design->getIteration()
            ]
        );
    }
}
```

I have started a new project, and looked back how I handled this before. I have used custom namers, property namers with a custom getter method (which isn't really a property :)), and it was a bit of a mess. I wanted a nice uniform way in the new project, so I made this.

The reason I'm not sure this belongs here is because you technically can do this with the property namer.

garak:
I'm closing this since it's a bit old and maybe a bit too specific.
Likely you can solve the problem just with a custom namer

PR #898, opened 2018-08-31
If form component is disabled then inject null instead of form's property accessor

norkunas:
In API we don't use form component at all so it's disabled completely but bundle tries to inject it's property accessor and fails. PropertyDirectoryNamer constructors allows to pass null so this works


garak:
Thanks!

PR #901, opened 2018-09-17
Allow overwriting of existing files when using Flysystem storage

### LINKED ISSUE #900: FlysystemStorage cannot overwrite existing file
|    Q        |   A
|------------ | -----
| Version     | 1.6.2

When using the FlysystemStorage, if you upload a document that has the same name as one that has previously been uploaded, the Flysystem package throws an exception "File already exists at path: XXXXX".

This is problematic when users are uploading a doc, then making changes, and then attempting to re-upload the same (now modified) document, with the same file name. The exception is thrown, causing the Flysystem storage to actually hold the old, out of date version of the document.

It looks like this is due to `$fs->writeStream()` being called, instead of `$fs->putStream()` on line 44 of `Vich\UploaderBundle\Storage\FlysystemStorage`

https://github.com/thephpleague/flysystem/issues/733#issuecomment-262880971
> `put()` doesn't append. It writes whether the file exists or not. `write()` only writes if the file does not exist, and `update()` only writes if the file exists.

Is there a reason why writeStream is used instead of putStream? 

What is the easiest way to override this behaviour if this is not considered a bug?

Thanks

benr77:
Allow overwriting of existing files when using Flysystem storage. Simply done by replacing writeStream() with putStream(). Fixes #900

### DIFF
--- a/Tests/Storage/FlysystemStorageTest.php
+++ b/Tests/Storage/FlysystemStorageTest.php
@@ -84,7 +84,7 @@ class FlysystemStorageTest extends StorageTestCase
 
         $this->filesystem
             ->expects($this->once())
-            ->method('writeStream')
+            ->method('putStream')
             ->with(
                 'originalName.txt',
                 $this->isType('resource'),


garak:
Thank you!

PR #907, opened 2018-10-05
Ability to display object in DownloadHandler

dehy:
This PR brings the ability to display the object (with `Content-Disposition: inline`) in the browser instead of forcing to download. In my project, I had the requirement of access control of the resource (so not exposing the file publicly) but also just displaying it in the browser (like PDFs).

This PR keeps retro-compatibility with previous behaviour of forced download.

Is this some behaviour you're willing to accept? If yes, I'll continue making tests and doc/

TODO:

- Update Test
- Update doc


garak:
LGTM.
You can go ahead

dehy:
Doc and Test added

PR #908, opened 2018-10-06
Update UploadHandler.php

Big-Shark:
Doesn't matter this is upload file, or file, we need just save
Example
```
$filePath = '/tmp/'.uniqid();
file_put_contents($filePath, file_get_contents('https://www.google.com/images/branding/googlelogo/2x/googlelogo_color_120x44dp.png'));
$entity->setImageFile(new File($filePath));
```

### DIFF
--- a/Handler/UploadHandler.php
+++ b/Handler/UploadHandler.php
@@ -3,7 +3,7 @@
 namespace Vich\UploaderBundle\Handler;
 
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
-use Symfony\Component\HttpFoundation\File\UploadedFile;
+use Symfony\Component\HttpFoundation\File\File;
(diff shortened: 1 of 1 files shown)


garak:
Can't you just use `new Uploaded file()`?

Big-Shark:
> Can't you just use `new Uploaded file()`?

No, because UploadedFile use php function is_uploaded_file(), and I receive validation error.
P.S. Yes, I can use 5th parameters in UploadedFile constructor, but this is look like a dirty.

garak:
It looks like you already solved your problem

Big-Shark:
> It looks like you already solved your problem

yes, but I think I used dirty hack, and I think this PR is better for all. If you have opposite opinion, please write me answer and close this PR.

PR #918, opened 2018-11-10
WIP: Upgrade to jms/metadata 2.0

stephanvierkant:
I'd like to upgrade to `jms/serializer` 2.0 which needs `jms/metadata` 2.0. I've tried to upgrade this library, but the test fails.

Any ideas how to fix the test?


stephanvierkant:
@garak the problem is "typehinting everywhere". That makes extending ‘ClassMetadata‘ impossible.

garak:
@stephanvierkant here or in jms/metadata?

stephanvierkant:
With JMS/metadata `ClassMetadata` can't be extended because it will violate return type hinting. Maybe that's a bad practice on JMS' side (maybe they should have type hinted an interface instead?), but I'm not sure about that.

But as far as I understand the code (didn't spend that much time yet, to be honest) it's not a problem for the library itself, only for the tests.

garak:
I'm not sure that type hinting is involved here, I see that our test is just setting a public property `$fields` and then trying to access it. When accessing `$fields`, test is failing because it can't find such property.

stephanvierkant:
In my PR, I've removed `Vich\UploaderBundle\Metadata\ClassMetadata`. This is because `AbstractFileDriver::loadMetadataFromFile` expects `Metadata\ClassMetadata`.

garak:
Is that removal needed? `Vich\UploaderBundle\Metadata\ClassMetadata` extends `Metadata\ClassMetadata`, so any expectation should be met

garak:
@stephanvierkant JMS published a document about upgrading to 2.0 in related issue

garak:
I see.
Well, without our `ClassMetadata`, there is no `$fields` anymore. I guess that this a crucial detail, because without that property the whole mapping is not working.

ikerib:
I have the same issue... is that fixed? how? I can't install VichUpload...

stephanvierkant:
> I have the same issue... is that fixed? how? I can't install VichUpload...

No, I'm waiting for an answer here: https://github.com/schmittjoh/metadata/issues/75#issuecomment-438427091

haterecoil:
Hello !

Thank you for the hard work!

In the meantime, I got on this thread since I had trouble setting up JMSSerializerBundle + VichUploaderBundle + FOSRestBundle in my project, I managed to install it by removing those dependencies and installing : 

```json
// composer.json
{
  "require": {
    "friendsofsymfony/rest-bundle": "^2.4.0",
    "jms/serializer-bundle": "^2.4.2",
    "vich/uploader-bundle": "^1.8",
  }
}
```

I will update everything once you manage to solve it. Keep it up 👍

stephanvierkant:
Still no update from JMS. Their serializer is great, but their support is terrible. Maybe we should get rid of it?

goetas:
> Their serializer is great, but their support is terrible. Maybe we should get rid of it?

Nice to see that, the response time for the first issue was less that 24h. :)
image

garak:
@stephanvierkant I understand your frustration, but please do not blame other libraries' maintainers: they're doing their best, just as we do.
Said that, can we try to avoid changing signature of `*::loadMetadataFromFile`, as suggested in related issue?

DGCarramona:
If I want to help with that pull request, how can I contribute to its code ?

garak:
Thanks for your interest @DGCarramona 
As mentioned above, we can try to follow suggestion given in [related issue](https://github.com/schmittjoh/metadata/issues/76#issuecomment-450802952)

ikerib:
is there any way to solve this issue and have this bundle installed correctly?

juntereiner:
I've been playing around with this PR and found a couple of things:

1. `Vich\UploaderBundle\Metadata\ClassMetadata` is mandatory to provide the $fields property.
There is no reason to change the signature return type of `::loadMetadataFromFile()` by `Vich\UploaderBundle\Metadata\ClassMetadata` which extends `Metadata\ClassMetadata` , even though it should be working on php7.3 with the variance return type (liskov).

2. I see you did `"jms/metadata": "^1.6|^2.0"` for the compatibility with php 7.1 I guess ? 
It won't work because `::loadMetadataFromFile()` on jms/metadata 1.6 has no return type and this PR add one.

So the only possibility is to support only jms/metadata 2.0 and raise the minimum php version to 7.2.
That's a big bc break and you should probably release a version 2.0 of this library. If you look at https://github.com/willdurand/Hateoas, that's exactly what he did.

I have a working solution if you want.

garak:
Thanks for your contribution @juntereiner
I look forward for your solution. Raising requirements to ^7.2 is not a problem

garak:
Closing in favor of #960 
Thank you anyway @stephanvierkant

PR #920, opened 2018-11-14
use default timezone UTC for phpunit tests

wuestkamp:
Hi there,

running the `./runTests.sh` on a machine where the php.ini has a different timezone set than UTC, for example 'America/Los_Angeles', fails some phpunit tests.

How about we set the timezone to default UTC for running the testsuite?

Cheers,
Kim


garak:
thank you @wuestkamp

Issue #929, opened 2018-11-24
CurrentDateTimeDirectoryNamer wrong date calculation


garak:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 1.8.4

#### Summary

Originally reported in https://github.com/dustin10/VichUploaderBundle/commit/ee11ae5883ce2178aefee0bc6d615889f6a56253#r31427035

#### Current behavior

When URL is generated for an already uploaded file, `CurrentDateTimeDirectoryNamer` gets a directory name based on current timestamp

#### How to reproduce

Setup `CurrentDateTimeDirectoryNamer`, upload a file. Then, try to call the same file on a different date (for example, the day after).

#### Expected behavior

`CurrentDateTimeDirectoryNamer` should get a directory name based on actual date of upload

#### Proposed solution

Add a new option to the namer, with a way to retrieve date of upload (a property name, maybe?).
To avoid BC break, such option must be not mandatory (with a fallback on current behavior).
We also need to update documentation accordingly

PR #930, opened 2018-12-01
Try to fix issue #929

garak:


dimonx3:
@garak 
I have this error. 'An exception has been thrown during the rendering of a template ("Warning: date() expects parameter 2 to be integer, object given").'
Method `CurrentDateTimeDirectoryNamer::directoryName()` requires small fix.
Need to call a function DateTime::getTimestamp(). Because second parameter `$timestamp` in function `date( string $format [, int $timestamp = time() ] )` must be integer type. But you pass the `DateTime` type.
Replace
`$dateTime = $this->propertyAccessor->getValue($object, $this->dateTimeProperty);`
to
`$dateTime = $this->propertyAccessor->getValue($object, $this->dateTimeProperty)->getTimestamp();`

garak:
Or: you can make your method return a timestamp

dimonx3:
But the property is named `date_time_property`. Doctrine also maps all date types (date, time, datetime, datetimetz) to a PHP DateTime object.
Therefore, it would be more correct to pass an `DateTime object` rather than a `int timestamp` to `date_time_property` property.
@garak

PR #958, opened 2019-01-31
add Content-Length header to DownloadHandler

abrgitlab:


garak:
So, this is the example from docs:
```php
class AcmeController extends AbstractController
{
    public function downloadImageAction(Image $image, DownloadHandler $downloadHandler): Response
    {
        return $downloadHandler->downloadObject($image, $fileField = 'imageFile');
    }
}
```
can't you just do like so?
```php
class AcmeController extends AbstractController
{
    public function downloadImageAction(Image $image, DownloadHandler $downloadHandler): Response
    {
        $fstat = fstat($imgage);
        $response = $downloadHandler->downloadObject($image, $fileField = 'imageFile');
        $response->headers->set('Content-Length', $fstat['size']);

        return $response;
    }
}
```

garak:
So, can't you just move fstat one line below?

garak:
OK, I got a bit deeper and I see that you can't use fstat outside DownloadHandler.
Anyway, your proposed solution is not working, see failing tests

garak:
DownloadHelper is using a StreamedResponse, not a BinaryResponse, so you can't calculate the length. Moreover, you cat only use fstat on local files, while this bundle supports a filesystem abstraction.

PR #960, opened 2019-02-16
Upgrade jms/metadata to 2.0

juntereiner:
Hello,

This is the continuation of #918.
PR also contains:
- Up minimum php version to 7.2
- Used ::class instead of FQCN string
- phpcsfixer fix
- remove php 7.1 from travis
- fix some tests deprecations

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,9 +20,10 @@
         }
     },
     "require": {
-        "php": "^7.1",
+        "php": "^7.2",
+        "ext-simplexml": "*",
         "behat/transliterator": "^1.2",
-        "jms/metadata": "^1.6",
+        "jms/metadata": "^2.0",
         "symfony/config": "^3.4|^4.0",
         "symfony/dependency-injection": "^3.4|^4.0",
         "symfony/event-dispatcher": "^3.4|^4.0",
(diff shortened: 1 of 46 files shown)


composer.json
```diff
@@ -20,9 +20,10 @@
         }
     },
     "require": {
-        "php": "^7.1",
+        "php": "^7.2",
+        "ext-simplexml": "*",
         "behat/transliterator": "^1.2",
-        "jms/metadata": "^1.6",
+        "jms/metadata": "^2.0",
```
guilliamxavier, line 26: The changelog for https://github.com/dustin10/VichUploaderBundle/releases/tag/1.9.0 says "allow for `jms/metadata` version 2", but that's not *allow for*, that's *require*! As a result, I can't install `vich/uploader-bundle` ^1.13 (needed for `league/flysystem-bundle` support) in a project that has a transitive dependency on `jms/metadata` 1.7 😢

I don't understand https://github.com/dustin10/VichUploaderBundle/pull/918#issuecomment-464262678

> 2. I see you did `"jms/metadata": "^1.6|^2.0"` for the compatibility with php 7.1 I guess ?
>    It won't work because `::loadMetadataFromFile()` on jms/metadata 1.6 has no return type and this PR add one.

That argument seems invalid: it *is* legal to add a return type when implementing/overriding (https://3v4l.org/rq2pd) (it isn't to remove, https://3v4l.org/oMqN5).

Would you accept a PR that changes `"^2.0"` to `"^1.6 || ^2.0"` (if the CI passes)?

garak: Of course! Thank you for noticing that

guilliamxavier: Thanks, see #1212

Issue #961, opened 2019-02-18
An error occurred while loading the web debug toolbar after upgrade to 1.9.0 


diabl0:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 1.9.0

#### Summary

While using latest Symfony 4.2.3 after upgrading to 1.9.0 web debug toolbbar stops working:
`An error occurred while loading the web debug toolbar.`

#### Current behavior

After upgrading to 1.9.0 web debug toolbbar stops working:
`An error occurred while loading the web debug toolbar.`

while accessing _wdt directly i have exception:

`The profiler template "VichUploaderBundle::Collector/mapping_collector.html.twig" for data collector "vich_uploader.mapping_collector" does not exist.`
at vendor/symfony/web-profiler-bundle/Profiler/TemplateManager.php line 90

Same happens when trying to access profiler directly

#### How to reproduce

Upgrade bundle from 1.8.6 to 1.9.0

#### Expected behavior

working web debug toolbar

garak:
That's weird, since the file is [present](https://github.com/dustin10/VichUploaderBundle/blob/1.9.0/Resources/views/Collector/mapping_collector.html.twig).
Can you find such file in your vendor dir?

PR #962, opened 2019-02-19
Fix issue with not loading the profiler template

dragosprotung:
Possible fix for #961.

I am not sure about the implications about this change, nor how to cover it with a test case. But it seems to solve the profiler toolbar issue under Symfony 4.2


garak:
Thank you @dragosprotung

pacid:
It was still an issue for me, having latest VichUploader and SF 4.4.1. Had to uninstall VichUploader because I am currently not doing anything with VichUploader, but will probably revisit this :)

PR #923, opened 2018-11-15
Downloader handler mimeType guessing

k20human:
---
name: Downloader handler mimeType guessing
about: Try to use file mimeType if it's already stored in object
---

### Improvement

|    Q        |   A
|------------ | ------
| New Feature | no
| RFC         | no
| BC Break    | no

#### Summary

Try to use file mimeType if it's already stored in object
Otherwise use $file->getMimeType

It's usefull if mimetype is alread stored in database for exemple and use files store with Flystsem. (In my case Google Cloud Storage)

Possible fix of #841

PR #965, opened 2019-02-27
Fix #774 and covers #923

### LINKED ISSUE #774: $file in downloadhelper is always null,
Hello,

I try to serve a file by a controller, but type Mime is always application/octet-stream

After xdebug the solution, i found the $file attribute of downloadHandler is always null

```
if (true === $fileName) {
            $fileName = $mapping->readProperty($object, 'originalName');
        }

        $file = $mapping->getFile($object);
```

My entity : 

```
<?php

namespace AppBundle\Entity;

use Doctrine\ORM\Mapping as ORM;
use Symfony\Component\HttpFoundation\File\File;
use Symfony\Component\HttpFoundation\File\UploadedFile;
use Vich\UploaderBundle\Entity\File as EmbeddedFile;
use Vich\UploaderBundle\Mapping\Annotation as Vich;

/**
 * Candidature
 *
 * @ORM\Table(name="candidature")
 * @ORM\Entity(repositoryClass="AppBundle\Repository\CandidatureRepository")
 * @Vich\Uploadable
 */
class Candidature
{
    /**
     * @var int
     *
     * @ORM\Column(name="id", type="integer")
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    private $id;

    /**
     * @var \DateTime
     *
     * @ORM\Column(name="date_updated", type="datetime")
     */
    private $dateUpdated;

    /**
     * NOTE: This is not a mapped field of entity metadata, just a simple property.
     * 
     * @Vich\UploadableField(mapping="candidature_file", fileNameProperty="file.name", size="file.size", mimeType="file.mimeType", originalName="file.originalName")
     * 
     * @var File
     */
    private $fileFile;

    /**
     * @ORM\Embedded(class="Vich\UploaderBundle\Entity\File")
     *
     * @var EmbeddedFile
     */
    private $file;



    public function __construct(){
        $this->dateUpdated = new \DateTime('NOW');
        $this->file = new EmbeddedFile();
    }


    /**
     * Get id
     *
     * @return integer 
     */
    public function getId()
    {
        return $this->id;
    }

    /**
     * If manually uploading a file (i.e. not using Symfony Form) ensure an instance
     * of 'UploadedFile' is injected into this setter to trigger the  update. If this
     * during Doctrine hydration.
     *
     * @param File|UploadedFile $file
     */
    public function setFileFile(File $file = null)
    {
        $this->fileFile = $file;

        if ($file) {
            // It is required that at least one field changes if you are using doctrine
            // otherwise the event listeners won't be called and the file is lost
            $this->dateUpdated = new \DateTimeImmutable();
        }

        return $this;
    }

    /**
     * @return File|null
     */
    public function getFileFile()
    {
        return $this->fileFile;
    }

    /**
     * @param EmbeddedFile $image
     */
    public function setFile(EmbeddedFile $file)
    {
        $this->file = $file;

        return $this;
    }

    /**
     * @return EmbeddedFile
     */
    public function getFile()
    {
        return $this->file;
    }
}
```

My controller

```
/** 
     * Serves an uploaded file.
     *
     * @Route("/{id}/file", name="candidature_file")
     */
    public function fileAction(Candidature $candidature)
    {   
        $downloadHandler = $this->get('vich_uploader.download_handler');

        return $downloadHandler->downloadObject($candidature, $fileField = 'fileFile', Candidature::class, true);
    } 
```

mysiar:
### Improvement

Fix: #774 and also covers #923 

|    Q        |   A
|------------ | ------
| New Feature | no
| RFC         | no
| BC Break    | no

#### Summary
Fixes null file and exception thrown during `$mapping->getFile($object)` for mimeType

### DIFF
--- a/Handler/DownloadHandler.php
+++ b/Handler/DownloadHandler.php
@@ -2,6 +2,7 @@
 
 namespace Vich\UploaderBundle\Handler;
 
+use Symfony\Component\HttpFoundation\File\Exception\FileNotFoundException;
 use Symfony\Component\HttpFoundation\ResponseHeaderBag;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 use Vich\UploaderBundle\Exception\NoFileFoundException;
@@ -40,12 +41,23 @@ class DownloadHandler extends AbstractHandler
             $fileName = $mapping->readProperty($object, 'originalName');
         }
 
-        $file = $mapping->getFile($object);
+        $mimeType = $mapping->readProperty($object, 'mimeType');
+
+        if (null === $mimeType) {
+            try {
+                $file = $mapping->getFile($object);
+                if (null !== $file) {
+                    $mimeType = $file->getMimeType();
+                }
+            } catch (FileNotFoundException $exception) {
+                $mimeType = null;
+            }
+        }
 
         return $this->createDownloadResponse(
             $stream,
             $fileName ?: $mapping->getFileName($object),
-            null === $file ? null : $file->getMimeType(),
+            $mimeType,
             $forceDownload
         );
     }
--- a/Tests/Handler/DownloadHandlerTest.php
+++ b/Tests/Handler/DownloadHandlerTest.php
@@ -160,10 +160,12 @@ class DownloadHandlerTest extends TestCase
         $this->object->setImageOriginalName('original-name.jpeg');
 
         $this->mapping
-            ->expects($this->once())
+            ->expects($this->exactly(2))
             ->method('readProperty')
-            ->with($this->object, 'originalName')
-            ->will($this->returnValue($this->object->getImageOriginalName()));
+            ->will($this->returnValueMap([
+                [$this->object, 'originalName', $this->object->getImageOriginalName()],
+                [$this->object, 'mimeType', $this->object->getImageMimeType()],
+            ]));
 
         $file = $this->getUploadedFileMock();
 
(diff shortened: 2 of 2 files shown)


Handler/DownloadHandler.php
```diff
@@ -40,12 +41,21 @@ public function downloadObject($object, string $field, ?string $className = null
             $fileName = $mapping->readProperty($object, 'originalName');
         }
 
-        $file = $mapping->getFile($object);
+        $mimeType = $mapping->readProperty($object, 'mimeType');
+
+        if (!$mimeType) {
```
garak: Can we avoid implicit casting?

Handler/DownloadHandler.php
```diff
@@ -40,12 +41,21 @@ public function downloadObject($object, string $field, ?string $className = null
...
+            try {
+                $file = $mapping->getFile($object);
+                $file ? $mimeType = $file->getMimeType() : null;
```
garak: Same here

mysiar: @garak please elaborate, I do not know what you mean

garak: I mean same as noted above: avoid implicit casting

mysiar: @garak done

garak:
Thank you @mysiar

PR #966, opened 2019-03-13
Accept entity inside Transliterator

Grafikart:
If an entity is linked to another the namer should be able to handle entities

```yaml
mappings:
    post_upload:
        service: vich_uploader.namer_directory_property
        options: { property: 'category', transliterate: true}
```


Grafikart:
wrong file

PR #967, opened 2019-03-13
Allow entity property for directory namer

Grafikart:
For a project I have to "group" files using a directory. I created a custom namer but I think it could be nice for VichUploaderBundle to be able to handle entity for the directory namer:

```yaml
vich_uploader:
    db_driver: orm

    mappings:
        blog_upload:
            upload_destination: upload
            directory_namer:
                service: App\Uploader\PropertyDirectoryNamer
                options: { property: 'category' }
```

if `getCategory()` returns an entity the namer will look for a `getId()` method instead of throwing an error.

### DIFF
--- a/Naming/PropertyDirectoryNamer.php
+++ b/Naming/PropertyDirectoryNamer.php
@@ -69,6 +69,10 @@ class PropertyDirectoryNamer implements DirectoryNamerInterface, ConfigurableInt
             throw new NameGenerationException(sprintf('Directory name could not be generated: property %s is empty.', $this->propertyPath));
         }
 
+        if (is_object($name) && method_exists($name, 'getId')) {
+            $name = $name->getId();
+        }
+
         if ($this->transliterate) {
             $name = Transliterator::transliterate($name);
         }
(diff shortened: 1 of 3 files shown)


garak:
I'm not sure to understand your use case. 
Also, it looks like it's a bit specific... for example, why the object is supposed to return its name via a `getId` method?

Grafikart:
For instance we could imagine that an `Attachment` is linked to a `Document`. 

I want to group `Attachment` using the `Document` id attached to it.

```
attachments/1/a.pdf
attachments/1/b.pdf
attachments/1/c.pdf
attachments/2/a.pdf
attachments/2/a.pdf
attachments/{documentID}/a.pdf
```

With this PR we can use the `getDocument` to retrieve the ID of the document instead of having to create a new getter `getDocumentId` just for the namer.

```yaml
                service: App\Uploader\PropertyDirectoryNamer
                options: { property: 'document' }
```

If you think it's too specific, feel free to close this, since if can be solved using a custom Namer.

garak:
Well, I'm afraid that it's too specific, I understand that you're using ID as name, but that's not a general rule. So maybe it's better to leave this in a custom namer.
Thank you anyway for your contribution. Feel free to propose other ideas, in case.

Issue #968, opened 2019-03-13
Make bundle compatible to twig/twig:^2.7.0


Lyhl:
Due to a security issue in twig/twig:2.6.2 and earlier versions, we had to update to twig/twig:^2.7.0 - see https://symfony.com/blog/twig-sandbox-information-disclosure for more details. 

Unfortunately after updating to twig 2.7.2 we see deprecation warnings caused by using this bundle. 

Can you please fix the deprecation warnings and provide a version that's fully compatible with twig/twig:^2.7.0

garak:
Can you report you deprecations?
I just tried to upgrade a project of mine, I can't find any twig-related deprecations caused by VichUploaderBundle

Lyhl:
Thank you for your quick answer. 

The deprecation notice "The spaceless tag is deprecated since Twig 2.7, use the spaceless filter instead." occurs on rendering a form using the VichImageType. 
I guess that in https://github.com/dustin10/VichUploaderBundle/blob/master/Resources/views/Form/fields.html.twig is the responsible code / spaceless tag. 
Using {% filter spaceless %} instead should solve the deprecation as mentioned in https://twig.symfony.com/doc/2.x/deprecated.html#tags. Please check if there are any other deprecations that should be solved in VichUploaderBundle.

Thx and regards,
Andreas

PR #969, opened 2019-03-17
Use whitespace control tags instead of spaceless

jbafford:
This allows the whitespace removal to be done at compile time, rather than runtime (thus improving performance).

Additionally, this will be more compatible, as the spaceless filter does not exist until Twig 2.7 (and composer.json does not declaring any particular version of Twig as a requirement).

This is a strict improvement to #968.

### DIFF
--- a/Resources/views/Form/fields.html.twig
+++ b/Resources/views/Form/fields.html.twig
@@ -1,47 +1,43 @@
-{% block vich_file_row -%}
-    {% set force_error = true %}
-    {{- block('form_row') }}
-{%- endblock %}
+{%- block vich_file_row -%}
+    {%- set force_error = true -%}
+    {{- block('form_row') -}}
+{%- endblock -%}
 
-{% block vich_file_widget %}
-    {% filter spaceless %}
-        <div class="vich-file">
-            {{ form_widget(form.file) }}
-            {% if form.delete is defined %}
-                {{ form_row(form.delete) }}
-            {% endif %}
+{%- block vich_file_widget -%}
+    <div class="vich-file">
+        {{- form_widget(form.file) -}}
+        {%- if form.delete is defined -%}
+            {{- form_row(form.delete) -}}
+        {%- endif -%}
 
-            {% if download_uri %}
-                <a href="{{ download_uri }}">{{ translation_domain is same as(false) ? download_label : download_label|trans({}, translation_domain) }}</a>
-            {% endif %}
-        </div>
-    {% endfilter  %}
-{% endblock %}
+        {%- if download_uri -%}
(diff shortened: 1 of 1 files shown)


garak:
Indeed, the spaceless filter was added in [Twig 1.38](https://twig.symfony.com/doc/1.x/filters/spaceless.html)

Resources/views/Form/fields.html.twig
```diff
@@ -3,45 +3,41 @@
...
 {%- endblock %}
 
-{% block vich_file_widget %}
-    {% filter spaceless %}
-        <div class="vich-file">
-            {{ form_widget(form.file) }}
-            {% if form.delete is defined %}
-                {{ form_row(form.delete) }}
-            {% endif %}
+{%- block vich_file_widget -%}
```
garak: I guess that starting space removal is useless when opening block.
Same applies for ending space removal when closing blocks.

jbafford: I suppose it’s not strictly necessary, however, at least in my projects, I prefer to use space removal on all tags for consistency’s sake, and to reduce the likelihood of surprises down the road.

jbafford: However I can trim the spaceless tags to the minimum necessary, if you want.

garak: I'll appreciate consistency.

So, I see 3 possible options:
- space removal everywhere
- space removal everywhere except where blatantly unnecessary (as mentioned before)
- keep spaceless filter

jbafford:
Just pointing out that while Twig 1.38 did also add the spaceless filter, that was also released at the same time 2.7 was (on March 12, 2019). So, strictly speaking, if the spaceless filter were kept, the bundle should add ^1.38|^2.7 as the twig version requirement.

garak:
> Just pointing out that while Twig 1.38 did also add the spaceless filter, that was also released at the same time 2.7 was (on March 12, 2019). So, strictly speaking, if the spaceless filter were kept, the bundle should add ^1.38|^2.7 as the twig version requirement.

Good point.
Do you mind to add requirement to this PR?

Resources/views/Form/fields.html.twig
jbafford: The changes I committed do space removal everywhere; the only tags that don’t were the open/close tags for vich_file_row, vich_image_row, and vich_file_label, which didn’t strictly need them, and I left them off to reduce the diff size. So, I’ll change them to use space removal so the whole file is consistent.

jbafford:
Since the PR removes the use of the spaceless filter entirely, we don’t need to add that requirement.

garak:
Thank you

PR #974, opened 2019-03-18
Urlencode the file name

gnat42:
When a file has characters in it that require escaping for urls we need to escape them.

### DIFF
--- a/Storage/FileSystemStorage.php
+++ b/Storage/FileSystemStorage.php
@@ -48,7 +48,7 @@ class FileSystemStorage extends AbstractStorage
         $uploadDir = $this->convertWindowsDirectorySeparator($mapping->getUploadDir($obj));
         $uploadDir = empty($uploadDir) ? '' : $uploadDir.'/';
 
-        return sprintf('%s/%s', $mapping->getUriPrefix(), $uploadDir.$name);
+        return sprintf('%s/%s', $mapping->getUriPrefix(), $uploadDir.urlencode($name));
     }
 
     private function convertWindowsDirectorySeparator(string $string): string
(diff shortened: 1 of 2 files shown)


Storage/FileSystemStorage.php
```diff
@@ -48,7 +48,7 @@ public function resolveUri($obj, string $mappingName, string $className = null):
         $uploadDir = $this->convertWindowsDirectorySeparator($mapping->getUploadDir($obj));
         $uploadDir = empty($uploadDir) ? '' : $uploadDir.'/';
 
-        return sprintf('%s/%s', $mapping->getUriPrefix(), $uploadDir.$name);
+        return sprintf('%s/%s', $mapping->getUriPrefix(), $uploadDir.urlencode($name));
```
garak: I'm afraid that storage resolution of filename is not the proper place to resolve presentation problems.

gnat42: I 100% agree, however the twig helper gets the full path back so if you urlencode that its even worse. Do you have suggestions on a proper way to solve the issue?

garak: What about validating the file name?
Or, even more simpler, using a namer that do the conversion at upload time

gnat42: I was thinking about that - however just today I had an issue where an image was uploaded with a space in the file name. This really shouldn't be a problem but wasn't working. I don't know why this is suddenly an issue, however its somewhat of a burden on the developer to use the bundle if file names with url unsafe chars need to be escaped/transformed but it isn't mentioned. What happens if you upload a file with a space or % in your projects?

garak: Bundle provides HashNamer, UniqIdNamer, and other namers to avoid such problems. If you chose to use original namer, you need to take into accounts this kind of problems.
As already mentioned, it's trivial to solve it, just implement your namer that do urlencoding or other forms of escaping.

garak:
I have to reject this PR for the reasons mentioned above.
Thanks anyway for your contribution

PR #979, opened 2019-03-21
add the missing asset twig command

12rambau:
If you do not use "asset", the "vich_uploader_asset" return a relative webPath that of course (depending on your routing) point on nothing


garak:
I'm using this bundle since ages and I never had a problem with "vich_uploader_asset".
Double check your configuration, likely you missed some steps

12rambau:
This is a solution I found on the web, I will triple check (If I'm the only one I'm sure I misunderstand something)

PR #996, opened 2019-05-14
#793 - Allow the Bundle to search Uploadable annotation in Parent Classes

### LINKED ISSUE #793: Allow to Put @Vich\UploadableField Annotation into a Parent Class
Using the `@Vich\UploadableField` annotation in a parent class (`@ORM\MappedSuperclass`) does not work, since the annotation driver only checks the class itself for the `UploadableField` annotation.

So you have to clone the `UploadableField` and its methods into each derivation class.

The following modification in the AnnotationDriver.php loadMetadataForClass() helps:

old code: 

`foreach ($class->getProperties() as $property) {`

new code:

~~~
$classes = array();
do {
    $classes[] = $class;
    $class = $class->getParentClass();
} while (false !== $class);
$classes = array_reverse($classes, false);
$properties = array();
foreach ($classes as $class) {
    $properties = array_merge($properties, $class->getProperties());
}
foreach ($properties as $property) {
~~~
Note:
- loop borrowed from `vendor/jms/metadata/src/Metadata/MetadataFactory.php::getClassHierarchy`
- Actually the code can be optimized to avoid using two loops.
- I'm referring to version `1.4.2`, the code has changed on trunk on github but has still the  same behavior.

etshy:
Closes #793

This PR allow the bundle to search for the Uploadable Annotation in parent Classes.

Thanks to HSteeb for the code

### DIFF
--- a/Metadata/Driver/AnnotationDriver.php
+++ b/Metadata/Driver/AnnotationDriver.php
@@ -41,13 +41,13 @@ class AnnotationDriver implements AdvancedDriverInterface
         $classMetadata = new ClassMetadata($class->name);
         $classMetadata->fileResources[] = $class->getFileName();
     
-        $classes = array();
+        $classes = [];
         do {
             $classes[] = $class;
             $class = $class->getParentClass();
         } while (false !== $class);
         $classes = array_reverse($classes, false);
-        $properties = array();
+        $properties = [];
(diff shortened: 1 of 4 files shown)


etshy:
I completely messed up my commits, I'll redo that frow scratch soon.

PR #997, opened 2019-05-14
Allow Uploadable annotation in parent classes

### LINKED ISSUE #793: Allow to Put @Vich\UploadableField Annotation into a Parent Class

etshy:
Closes #793

This PR allow the bundle to search for the Uploadable Annotation in parent Classes.

Thanks to HSteeb for the code


Metadata/Driver/AnnotationDriver.php
```diff
@@ -19,36 +19,47 @@ class AnnotationDriver implements AdvancedDriverInterface
      * @deprecated
      */
     const UPLOADABLE_ANNOTATION = Uploadable::class;
-
+
```
garak: This is bad. Please refer to [StyleCI issues](https://github.styleci.io/analyses/zGAaOl) and fix

etshy: I guess that's the default behaviour in PHPStorm as I don't remember configuring that
I'll change that as soon as I can.

Tests/DummyFile.php
```diff
@@ -0,0 +1,126 @@
+<?php
+declare(strict_types=1);
```
etshy: I saw an error in [StyleCI](https://github.styleci.io/analyses/zGAaOl) about this line but I can't see what's wrong exactly.
(the - and + line seems to be exactly the same).
Should I delete this line ?

garak: It looks like there's an extra empty line after `declare` line. Anyway, I don't see a reason to declare strict types in such file

etshy: It's just creating by default with my phpstorm settings.
It's true  that in tests file it's not needed

etshy:
Hmm I'm not sure what to do about the last styleci error.
I tried to add or remove empty line/indents but it always fails.

garak:
> Hmm I'm not sure what to do about the last styleci error.
> I tried to add or remove empty line/indents but it always fails.

You're missing an empty line at the end of file

garak:
Thank you @etshy

PR #1000, opened 2019-05-16
add SmartUniqueNamer

garak:
Currently, this bundle has a namer for uniqueness that is too harsh (using simply uniqid) and a namer that tries to keep original name, that is too dumb (not transforming name and prepending uniqid).
This namer tries to get the best of both worlds, so it's named "smart" 😎

### DIFF
--- a/Resources/doc/namers.md
+++ b/Resources/doc/namers.md
@@ -17,9 +17,10 @@ At the moment there are several available namers:
   * `Vich\UploaderBundle\Naming\PropertyNamer`
   * `Vich\UploaderBundle\Naming\HashNamer`
   * `Vich\UploaderBundle\Naming\Base64Namer`
+  * `Vich\UploaderBundle\Naming\SmartUniqeNamer`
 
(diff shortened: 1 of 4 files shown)


Resources/doc/namers.md
```diff
@@ -35,6 +36,10 @@ hash `algorithm` and result `length` of the file
 You can specify the `length` of the random string. Using this namer, foo.jpg will be uploaded as something
 like 6FMNgvkdUs.jpg
 
+**SmartUniqeNamer** will rename your uploaded files appending a strong uniqueid to the original name, while 
+apllying a transliteration. Using this namer, a Strange name.jpg will be uploaded as something like
```
etshy: There is a  typing error here, I think. 
"apllying" => "applying"

gnutix:
@garak That one is awesome! Any chance to get a release anytime soon ? (currently it's only in master)

garak:
@gnutix here you are! https://github.com/dustin10/VichUploaderBundle/releases/tag/1.10.0

PR #1011, opened 2019-06-17
Fix broken links

yordv:
Fix broken links for PSR-1 and PSR-2

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,8 +6,8 @@ First of all, **thank you** for contributing, **you are awesome**!
 Here are a few rules to follow in order to ease code reviews, and discussions before
 maintainers accept and merge your work.
 
-You MUST follow the [PSR-1](http://www.php-fig.org/psr/1/) and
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -6,8 +6,8 @@ First of all, **thank you** for contributing, **you are awesome**!
 Here are a few rules to follow in order to ease code reviews, and discussions before
 maintainers accept and merge your work.
 
-You MUST follow the [PSR-1](http://www.php-fig.org/psr/1/) and
-[PSR-2](http://www.php-fig.org/psr/2/). If you don't know about any of them, you
+You MUST follow the [PSR-1](https://www.php-fig.org/psr/psr-1/) and
```
garak: What about omitting protocol?

yordv: Its redirect to https if you hit http

garak: Exactly. So we can link to `//www.php-fig.org/psr/psr-1/` instead of `https://www.php-fig.org/psr/psr-1/`

garak:
Thank you!

Issue #858, opened 2018-03-29
Option 'imagine_pattern' generates a wrong link


murtukov:
| Q                          | A
| -------------------------- | -----
| Bug report?                | yes
| Feature request?           | no
| BC Break report?           | no
| Symfony version            | 4.0.6
| VichUploaderBundle version | 1.8.2
| LiipImagineBundle version | 2.0.x-dev
| SonataAdminBundle version | 3.33.0

I am using VichUploaderBundle + LiipImagineBundle + SonataAdminBundle.

### config/packages/vich_uploader.yaml

```
vich_uploader:
    db_driver: orm

    mappings:
        user_img:
            uri_prefix:         /img/user
            upload_destination: "%kernel.root_dir%/../public/img/user"
            namer: Vich\UploaderBundle\Naming\HashNamer
            directory_namer: Vich\UploaderBundle\Naming\SubdirDirectoryNamer
            delete_on_update:   true
            delete_on_remove:   true
```

### config/packages/imagine.yaml
```
liip_imagine:
    resolvers:
        default:
            web_path: ~

    filter_sets:
        cache: ~
        my_thumb:
            quality: 75
            filters:
                thumbnail: { size : [120, 90], mode : outbound }
                background: { size : [122, 92], position : center, color : '#eee' }
```

If I use this code somewhere in a template:
`<img src="{{ vich_uploader_asset(object, 'imageFile') | imagine_filter('my_thumb') }}" />`
it works perfectly and generates a correct link to thumbnail, like this:
`http://example.com/media/cache/my_thumb/img/user/78/78c737520d6620f766d5a636da9afa13e89bf47f.jpg`

However if I use the VichImageType with the option 'imagine_pattern' it generates a link without uri_prefix, like this:
`http://example.com/media/cache/my_thumb/78/78c737520d6620f766d5a636da9afa13e89bf47f.jpg`

It is the same entity in both cases. 
I use the VichImageType in SonataAdminBundle, but since the official documentation says:

> When defining fields in your admin classes you can use any of the standard Symfony field types and configure them as you would normally

it should make no difference.

I found the "problem", which was located in the file `Vich\UploaderBundle\Form\Type\VichImageType`.
If you change the line **64** from this:
`$path = $this->storage->resolvePath($object, $form->getName(), null, true);`
to this:
`$path = $this->storage->resolveUri($object, $form->getName(), null, true);`
everything works fine.

Thanx

garak:
[...]

PR #1021, opened 2019-07-17
Multiple ways of resolving the object's storage path

gnutix:
Fixes https://github.com/dustin10/VichUploaderBundle/issues/858#issuecomment-511848521

### DIFF
--- a/Form/Type/VichImageType.php
+++ b/Form/Type/VichImageType.php
@@ -19,6 +19,12 @@ use Vich\UploaderBundle\Storage\StorageInterface;
  */
 class VichImageType extends VichFileType
 {
+    const STORAGE_RESOLVE_URI = 0;
+
+    const STORAGE_RESOLVE_PATH_ABSOLUTE = 1;
+
+    const STORAGE_RESOLVE_PATH_RELATIVE = 2;
+
     /**
      * @var CacheManager|null
      */
@@ -42,7 +48,17 @@ class VichImageType extends VichFileType
         $resolver->setDefaults([
             'image_uri' => true,
             'imagine_pattern' => null,
+            'storage_resolve_method' => static::STORAGE_RESOLVE_URI,
         ]);
+
+        $resolver->setAllowedValues(
+            'storage_resolve_method',
+            [
+                static::STORAGE_RESOLVE_URI,
+                static::STORAGE_RESOLVE_PATH_RELATIVE,
+                static::STORAGE_RESOLVE_PATH_ABSOLUTE,
+            ]
+        );
 
         $resolver->setAllowedTypes('image_uri', ['bool', 'string', 'callable']);
 
@@ -66,7 +82,7 @@ class VichImageType extends VichFileType
                     throw new \RuntimeException('LiipImagineBundle must be installed and configured for using "imagine_pattern" option.');
                 }
 
-                $path = $this->storage->resolveUri($object, $form->getName(), null);
+                $path = $this->resolvePath($options['storage_resolve_method'], $object, $form);
                 if (null !== $path) {
                     $view->vars['image_uri'] = $this->cacheManager->getBrowserPath($path, $options['imagine_pattern']);
                 }
@@ -89,5 +105,22 @@ class VichImageType extends VichFileType
     public function getBlockPrefix(): string
     {
         return 'vich_image';
+    }
+
+    protected function resolvePath($storageResolveMethod, $object, FormInterface $form)
+    {
+        if (static::STORAGE_RESOLVE_URI === $storageResolveMethod) {
+            return $this->storage->resolveUri($object, $form->getName());
+        }
+
+        if (static::STORAGE_RESOLVE_PATH_ABSOLUTE === $storageResolveMethod) {
+            return $this->storage->resolvePath($object, $form->getName());
+        }
+
+        if (static::STORAGE_RESOLVE_PATH_RELATIVE === $storageResolveMethod) {
+            return $this->storage->resolvePath($object, $form->getName(), null, true);
+        }
+
+        return null;
     }
 }
--- a/Form/Type/VichImageType.php
+++ b/Form/Type/VichImageType.php
@@ -19,11 +19,11 @@ use Vich\UploaderBundle\Storage\StorageInterface;
  */
 class VichImageType extends VichFileType
 {
-    const STORAGE_RESOLVE_URI = 0;
+    public const STORAGE_RESOLVE_URI = 0;
 
-    const STORAGE_RESOLVE_PATH_ABSOLUTE = 1;
+    public const STORAGE_RESOLVE_PATH_ABSOLUTE = 1;
 
-    const STORAGE_RESOLVE_PATH_RELATIVE = 2;
+    public const STORAGE_RESOLVE_PATH_RELATIVE = 2;
 
     /**
      * @var CacheManager|null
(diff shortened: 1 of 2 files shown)


gnutix:
I can't get the same error in local using `vendor/bin/phpunit`. Any idea how I can reproduce it ?

Form/Type/VichImageType.php
```diff
@@ -19,6 +19,12 @@
  */
 class VichImageType extends VichFileType
```
garak: Why on VichImageType? Don't you think that such change would fit on VichFileType too?

gnutix: This code is meant only to generate the value of `image_uri` (which is specific to `VichImageType`, not `VichFileType`). So it feels in the right place to me.

garak: I guess that the problem described in related issue can apply also to a generic file: the user want to display a link to file, and link href attribute is technically the same as image src attribute (a path, in the end)

garak: You're right, the option applies only to VichImageType

Form/Type/VichImageType.php
```diff
@@ -19,6 +19,12 @@
...
 {
+    const STORAGE_RESOLVE_URI = 0;
```
garak: can we make constants private?

Form/Type/VichImageType.php
```diff
@@ -90,4 +106,21 @@ public function getBlockPrefix(): string
     {
         return 'vich_image';
     }
+
+    protected function resolvePath($storageResolveMethod, $object, FormInterface $form)
```
garak: Can we make this method private?
Can we add a return type?

Tests/Form/Type/VichImageTypeTest.php
```diff
@@ -204,6 +213,45 @@ public function testLiipImagineBundleIntegration(): void
         $this->assertEquals($vars, $view->vars);
     }
 
+    public function getLiipImagineBundleIntegrationData(): array
+    {
+        $field = 'image';
+        $object = new Product();
+
+        return [
+            [
```
garak: can we use keys to describe provided cases?

Form/Type/VichImageType.php
gnutix: Sure! I was not sure about the PHP supported version, but as it's 7.2 it's all good.

gnutix: Having them public allows to make the configuration quite readable :

```php
$builder->add('file', VichImageType::class, [
    'storage_resolve_method' => VichImageType::STORAGE_RESOLVE_PATH_ABSOLUTE,
]);
```

Where having them private would force to use integers or pre-defined strings for configuration. Not as nice IMO.

Tests/Form/Type/VichImageTypeTest.php
gnutix: Done

Form/Type/VichImageType.php
gnutix: Done

gnutix:
Hmm... I just realized there's another way to accomplish what I needed : 

```php
<?php
use Liip\ImagineBundle\Imagine\Cache\CacheManager;

final class MyFormType extends AbstractType
{
    private $cacheManager;

    public function __construct(CacheManager $cacheManager)
    {
        $this->cacheManager = $cacheManager;
    }

    public function buildForm(FormBuilderInterface $builder, array $options): void
    {
        $builder->add('file', VichImageType::class, [
            'image_uri' => function ($object, $resolvedUri) {
                return $this->cacheManager->getBrowserPath($object->getFileName(), 'my_imagine_filter');
            },
        ]);
    }
}
```

That way, I bypass the "storage->resolveUri" call and use my "relative filename" to generate the path using LiipImagine's CacheManager directly.

I'm wondering if that couldn't be the simplest answer to this issue, instead of adding all this new code. WDYT @BlackWiCKED ? Could you please try it out and let us know if that works for your case ?

BlackWiCKED:
> I'm wondering if that couldn't be the simplest answer to this issue, instead of adding all this new code. WDYT @BlackWiCKED ? Could you please try it out and let us know if that works for your case ?

Thank you for your efforts. It works partially @gnutix, it's a good thing that I can access the cached resources this way, but it also generates a lot of extra (and not necessarily desirable) work on the user side. I have more entities, and inserting the same closure into each form type is not what I would prefer - plus I need to set the constructor function and use the CacheManager class in each of my form type classes.

I said it works partially, because I noticed some strange behavior too. It might be specific to my case only though - but it did not happen with version 1.9.2 so I think it's worth mentioning. When there is no image uploaded to my entity, I use a placeholder image - which is an asset image file in my source code. That file is not stored in the private S3 bucket, but on my own server. It still gets cached though, and that cached image from the public S3 bucket is displayed on the site the same way as regular uploaded images. Basically I have two loaders (filesystem_loader and flysystem_loader) merged together with a chain loader in liip imagine config. 

`{{ (file.image|default(asset('build/images/noimage.png')))|imagine_filter('thumbnail') }}`

Looks like the problem is that when **_file.image_** is not set, the closure is still called via "image_uri" property, and since **_$object->getImage()_** is empty, **_$this->cacheManager->getBrowserPath_** returns a non existing path - it's just the cache folder - without the filename. The missing file appears in the browser's network tab. It can be fixed with something like this:

```php
$builder->add('file', VichImageType::class, [
    'image_uri' => function ($object/*, $resolvedUri*/) {
        if (!empty($object->getImage())) {
            return $this->cacheManager->getBrowserPath($object->getImage(), 'thumbnail');
        }
        return false;
    }
]);
```

Anyways, my opinion is that it just adds more complexity to something that should not be there at all. Setting properties like "imagine_pattern" or "storage_resolve_method" is fine, but writing and duplicating actual code - just to make a 3rd party library work - is not what I would prefer.

Form/Type/VichImageType.php
```diff
@@ -90,4 +106,21 @@ public function getBlockPrefix(): string
     {
         return 'vich_image';
     }
+
+    private function resolvePath($storageResolveMethod, $object, FormInterface $form): ?string
```
garak: can we type-hint? E.g. `int $storageResolveMethod, object $object`)

Form/Type/VichImageType.php
```diff
@@ -42,8 +48,18 @@ public function configureOptions(OptionsResolver $resolver): void
...
+            [
+                static::STORAGE_RESOLVE_URI,
+                static::STORAGE_RESOLVE_PATH_RELATIVE,
+                static::STORAGE_RESOLVE_PATH_ABSOLUTE,
+            ]
+        );
+
         $resolver->setAllowedTypes('image_uri', ['bool', 'string', 'callable']);
 
         $imageUriNormalizer = function (Options $options, $imageUri) {
```
garak: We could add type hint and return type here, too

Form/Type/VichImageType.php
gnutix: I thought about it, but noticed that in `VichFileType` there was a similar method with no type hints. So I restrained, not knowing if there was a good reason for that. Could you confirm we want these type hints ? And should I add them in the other method too while I'm at it?

garak: I'd prefer such changes in files that need to be changed for other reasons (like this one). We'll handle other files when needed (or in a PR dedicated to bring consistency)

gnutix:
If someone wants to take over this PR, please do. I've removed LiipImagine from my project in the end and don't have more time to dedicate to this.

garak:
Closing in favor of #1045

PR #1026, opened 2019-08-30
Compatibility with doctrine/mongodb-odm v.2.0; Drop DoctrineClassUtils;

shadz3rg:
Trying to read metadata on a referenced document class (proxy) results in a critical exception.
Since Doctrine\Common\Util\ClassUtils is considered as deprecated:
https://github.com/doctrine/common/issues/867
I suggest this implementation:
https://github.com/api-platform/core/blob/master/src/Util/ClassInfoTrait.php#L38

### DIFF
--- a/Util/ClassUtils.php
+++ b/Util/ClassUtils.php
@@ -22,10 +22,25 @@ class ClassUtils
      */
     public static function getClass($object): string
     {
-        if (\class_exists(DoctrineClassUtils::class)) {
-            return DoctrineClassUtils::getClass($object);
+        $className = \get_class($object);
+
+        // __CG__: Doctrine Common Marker for Proxy (ODM < 2.0 and ORM < 3.0)
+        // __PM__: Ocramius Proxy Manager (ODM >= 2.0)
+        if ((false === $positionCg = strrpos($className, '\\__CG__\\')) &&
+            (false === $positionPm = strrpos($className, '\\__PM__\\'))) {
+            return $className;
         }
 
-        return \get_class($object);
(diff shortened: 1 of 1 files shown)


Util/ClassUtils.php
```diff
@@ -22,10 +20,25 @@ private function __construct()
      */
     public static function getClass($object): string
```
garak: What about type-casting `object` here?

Util/ClassUtils.php
```diff
@@ -22,10 +20,25 @@ private function __construct()
...
     {
-        if (\class_exists(DoctrineClassUtils::class)) {
-            return DoctrineClassUtils::getClass($object);
+        $className = \get_class($object);
+
+        // __CG__: Doctrine Common Marker for Proxy (ODM < 2.0 and ORM < 3.0)
```
garak: I think we should credit original source here

Util/ClassUtils.php
shadz3rg: @garak did you mean type-hinting? 
`public static function getClass(object $object): string`
or we should expect something different than original object here?

Util/ClassUtils.php
shadz3rg: @garak ofc

Util/ClassUtils.php
garak: We expect an object, as suggested by variable name and by using `get_class` on it. So, passing a non-object was already not valid, but without type hint could be unnoticed. I think it's better to  be explicit and to fail with an error in such case

PR #1028, opened 2019-09-01
Implement Dockerfile for local testing

p365labs:
Hi all I've tried to implement a Dockerfile for testing on local environment.
This will allow idempotent test on differents workstations with different OS
and fixing library versions (php & pecl & so on)

what do u think?


CONTRIBUTING.md
```diff
@@ -31,25 +31,38 @@ Thank you!
...
 
 ```bash
-SYMFONY_VERSION=3.4.1 ./runTests.sh --filter testCustomFileNameProperty
+make test TEST="Tests/Util/FilenameUtilsTest.php"
+```
+
+will allow to tests single Test Classes.
+
+There are 3 environments available: PHP 7.2, 7.3 and 7.4-rc.
+Default environment is *PHP 7.2* if you want to execute it against
```
garak: Is there a reason for 7.2 being default? I guess that 7.3 would be a better choice, being the latest stable version.

composer.json
```diff
@@ -47,6 +47,7 @@
         "symfony/framework-bundle": "^3.4.23 || ^4.2.4",
         "symfony/phpunit-bridge": "^4.3",
         "symfony/security-csrf": "^3.4 || ^4.0",
+        "symfony/symfony": "~3.4.0",
```
garak: Why this add to requirements?

CONTRIBUTING.md
p365labs: no absolutely no a real reason. u're right! going to update it.

composer.json
p365labs: my mistake... I forgot to remove it.

.gitattributes
```diff
@@ -11,3 +11,5 @@
 /Tests              export-ignore
 /Resources/meta     export-ignore
 /Resources/doc      export-ignore
+/docker             export-ignore
+/.Makefile          export-ignore
```
garak: There's a typo here: it should be `/Makefile`, not `/.Makefile`

p365labs: ahahahah my bad :( sorry @garak

garak:
Thank you @p365labs

PR #1029, opened 2019-09-03
Fix deprecation for using spaceless tag

botjaeger:
for symfony 3.4

### DIFF
--- a/Resources/views/Form/fields.html.twig
+++ b/Resources/views/Form/fields.html.twig
@@ -4,7 +4,7 @@
 {%- endblock %}
 
 {% block vich_file_widget %}
-    {% spaceless %}
+    {% apply spaceless %}
         <div class="vich-file">
             {{ form_widget(form.file) }}
             {% if form.delete is defined %}
@@ -15,7 +15,7 @@
                 <a href="{{ download_uri }}">{{ translation_domain is same as(false) ? download_label : download_label|trans({}, translation_domain) }}</a>
             {% endif %}
         </div>
-    {% endspaceless %}
+    {% endapply %}
 {% endblock %}
 
 {% block vich_image_row -%}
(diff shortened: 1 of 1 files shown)


Resources/views/Form/fields.html.twig
```diff
@@ -4,7 +4,7 @@
 {%- endblock %}
 
 {% block vich_file_widget %}
-    {% spaceless %}
+    {% apply spaceless %}
```
garak: We can't just replace `spaceless` with `apply spaceless`, since this bundle is not requiring any specific version of Twig. So, if some users use an old version of Twig, their code would break.

I see two possible solutions: 
* add a `conflict` section in our composer.json (hard)
* use the whitespace "minus" operator instead (easy)

botjaeger: from this

```
{% block vich_file_widget %}
    {% apply spaceless %}
        <div class="vich-file">
            {{ form_widget(form.file) }}
            {% if form.delete is defined %}
                {{ form_row(form.delete) }}
            {% endif %}

            {% if download_uri %}
                <a href="{{ download_uri }}">{{ translation_domain is same as(false) ? download_label : download_label|trans({}, translation_domain) }}</a>
            {% endif %}
        </div>
    {% endapply %}
{% endblock %}
```

to this ?

```
{%- block vich_file_widget -%}
    <div class="vich-file">
        {{ form_widget(form.file) }}
        {% if form.delete is defined %}
            {{ form_row(form.delete) }}
        {% endif %}

        {% if download_uri %}
            <a href="{{ download_uri }}">{{ translation_domain is same as(false) ? download_label : download_label|trans({}, translation_domain) }}</a>
        {% endif %}
    </div>
{%- endblock -%}
```

garak: That's not enough, you should use `{{-` and `-}}` too

garak: Moreover: also other `{%` should be changed to  `{%-`

botjaeger: alright, will update the PR

PR #1032, opened 2019-09-04
Fix naming files with ".0" extension

### LINKED ISSUE #1030: Extension ".0" is disregarded when naming files
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 1.10.1

#### Summary

When uploading files with an `.0` extension, the available namers ignore the extension and rename the file as if it had no extension at all or based on a mime-type guess.

#### Current behavior

new file name `5d6faa6b945d0819644602` (using UniqidNamer as example)

#### How to reproduce

Use this mapping:

```
vich_uploader:
    # ...
    mappings:
        product_image:
            upload_destination: product_image_fs
            namer: Vich\UploaderBundle\Naming\UniqueNamer
```
Upload a file `examplefile.0` as product_image.

Failing unit tests are provided by 6ebc7af29c50b07beca01148cfe51e0cfcfcdfeb .

#### Expected behavior

new file name `5d6faa6b945d0819644602.0`

tantegerda1:
Files with an extension ".0" typically occur as output of file split
operations and compression utilities. Although unusual, such extensions
are valid and should be processable by VichUploader.

Fixes #1030

### DIFF
--- a/Naming/Polyfill/FileExtensionTrait.php
+++ b/Naming/Polyfill/FileExtensionTrait.php
@@ -17,11 +17,11 @@ trait FileExtensionTrait
     {
         $originalName = $file->getClientOriginalName();
 
-        if ($extension = \pathinfo($originalName, PATHINFO_EXTENSION)) {
+        if ('' !== ($extension = \pathinfo($originalName, PATHINFO_EXTENSION))) {
             return $extension;
         }
 
-        if ($extension = $file->guessExtension()) {
+        if ('' !== ($extension = $file->guessExtension())) {
             return $extension;
         }
 
--- a/Tests/Naming/UniqidNamerTest.php
+++ b/Tests/Naming/UniqidNamerTest.php
@@ -21,6 +21,9 @@ class UniqidNamerTest extends TestCase
             ['lala.mp3',      'mpga',            '/[a-z0-9]{13}.mp3/'],
             ['lala',          'mpga',            '/[a-z0-9]{13}.mpga/'],
             ['lala',          null,              '/[a-z0-9]{13}/'],
+            ['lala.0',        null,              '/[a-z0-9]{13}\\.0/'],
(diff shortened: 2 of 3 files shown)


Naming/Polyfill/FileExtensionTrait.php
```diff
@@ -17,11 +17,11 @@ private function getExtension(UploadedFile $file): ?string
     {
         $originalName = $file->getClientOriginalName();
 
-        if ($extension = \pathinfo($originalName, PATHINFO_EXTENSION)) {
+        if ('' !== ($extension = \pathinfo($originalName, PATHINFO_EXTENSION))) {
```
garak: I'd rather check for `null`

tantegerda1: Hm… I'm unsure about that.

The [php docs](https://www.php.net/manual/en/function.pathinfo.php#refsect1-function.pathinfo-returnvalues) says 

> If options is present, returns a string containing the requested element. 

and testing a file without extension `var_dump(\pathinfo('foo/bar', PATHINFO_EXTENSION));"` shows `string(0) ""`.

garak: You're right. So what is the purpose of checking with `is_string` below?

Tests/Naming/UniqidNamerTest.php
```diff
@@ -21,6 +21,9 @@ public function fileDataProvider(): array
             ['lala.mp3',      'mpga',            '/[a-z0-9]{13}.mp3/'],
             ['lala',          'mpga',            '/[a-z0-9]{13}.mpga/'],
             ['lala',          null,              '/[a-z0-9]{13}/'],
+            ['lala.0',        null,              '/[a-z0-9]{13}\\.0/'],
```
garak: Why `lala.0` name is returning `null` extension? I thought it was supposed to return `.0`

garak:
Thank you @tantegerda1

PR #1045, opened 2019-10-22
Diff from gnutix's PR.

BlackWiCKED:
I used gnutix's PR ( #1021 ) as the base of this new one.

I have not made any changes in the logic, and I have not run automated tests. I tried it on my own project, by using the new **_storage_resolve_method_** variable, and it worked as earlier (in version 1.9).

I also extended the documentation, but it definitely needs a review.

### DIFF
--- a/Form/Type/VichImageType.php
+++ b/Form/Type/VichImageType.php
@@ -106,7 +106,7 @@ class VichImageType extends VichFileType
         return 'vich_image';
     }
 
-	private function resolvePath($storageResolveMethod, $object, FormInterface $form): ?string
+	private function resolvePath(int $storageResolveMethod, object $object, FormInterface $form): ?string
 	{
 		if (static::STORAGE_RESOLVE_URI === $storageResolveMethod) {
 			return $this->storage->resolveUri($object, $form->getName());
--- a/Form/Type/VichImageType.php
+++ b/Form/Type/VichImageType.php
@@ -19,10 +19,11 @@ use Vich\UploaderBundle\Storage\StorageInterface;
  */
 class VichImageType extends VichFileType
 {
+    public const STORAGE_RESOLVE_URI = 0;
 
-	public const STORAGE_RESOLVE_URI = 0;
-	public const STORAGE_RESOLVE_PATH_ABSOLUTE = 1;
-	public const STORAGE_RESOLVE_PATH_RELATIVE = 2;
+    public const STORAGE_RESOLVE_PATH_ABSOLUTE = 1;
+
+    public const STORAGE_RESOLVE_PATH_RELATIVE = 2;
 
     /**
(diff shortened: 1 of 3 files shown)


Form/Type/VichImageType.php
```diff
@@ -42,8 +47,18 @@ public function configureOptions(OptionsResolver $resolver): void
...
+		    [
+			    static::STORAGE_RESOLVE_URI,
+			    static::STORAGE_RESOLVE_PATH_RELATIVE,
+			    static::STORAGE_RESOLVE_PATH_ABSOLUTE,
+		    ]
+	    );
+
         $resolver->setAllowedTypes('image_uri', ['bool', 'string', 'callable']);
 
         $imageUriNormalizer = static function (Options $options, $imageUri) {
```
BlackWiCKED: @garak I'm not sure about the return type here. It can be a bool, a string or even a callable, however PHP does not support multiple return type declarations (yet) - or at least I'm not aware of such. "Mixed" is not an option. Some people suggest to use an Interface for this, but that seems to be an overkill to me...

What are your thoughts? We have a very similar normalizer function for **_VichFileType_** class. Maybe this type hinting job should be done in a separate PR to finally release the solution for the original problem.

Form/Type/VichImageType.php
```diff
@@ -91,4 +105,19 @@ public function getBlockPrefix(): string
     {
         return 'vich_image';
     }
+
+	private function resolvePath(int $storageResolveMethod, object $object, FormInterface $form): ?string
```
BlackWiCKED: Added type hint for $storageResolveMethod and $object.

Form/Type/VichImageType.php
garak: Missing return type for mixed cases is fine

BlackWiCKED:
> What about `VichFileType`? Method `resolveUri` is used there too.

I believe this is VichImageType specific, because it is triggered only when the LiipImagineBundle (imagine_pattern) is involved. My issue was exactly that; I had to send relative path to the cache resolver, because the absolute path was not publicly accessible, so it could not handle it properly. I'm aware that my case is somewhat unique, but the PR gives the users option to chose whatever fits for their needs best.

By the way, you already discussed this in one of the previous PR's conversations, and then agreed with the contributor that:

> You're right, the option applies only to VichImageType

https://github.com/dustin10/VichUploaderBundle/pull/1021/files/1a3b792b8fbfa5cd40e82670228e5f5f20c8e1a3#diff-61cf138b1eea6444653e4ae361447fbd

garak:
> By the way, you already discussed this in one of the previous PR's conversations, and then agreed with the contributor that:
> 
> > You're right, the option applies only to VichImageType

Ah, that feeling of déjà vu :smile:

Issue #1049, opened 2019-10-29
Smarter namer is not smart enough


garak:
We currently have [https://github.com/dustin10/VichUploaderBundle/blob/master/Naming/SmartUniqueNamer.php](/SmartUniqueNamer).

It could be smarter:

* it could lowercase extension (who wants files like "mypicture.JPG"?)
* it could check for name length. Since most filesystems accept a maximum length of 255 characters, if an user tries to upload a file with very long name, the resulting name can exceed such limit (since the namer is adding the uniqid prefix)

PR #1052, opened 2019-10-29
Sets to lowercase, smartly handles names over 255 characters

aewens:
In reference to issue #1049

### DIFF
--- a/Naming/SmartUniqueNamer.php
+++ b/Naming/SmartUniqueNamer.php
@@ -24,6 +24,7 @@ final class SmartUniqueNamer implements NamerInterface
         $uniqBasename = \sprintf('%s%s', $originalBasename, $uniqId);
         $uniqExtension = \sprintf('%s.%s', $uniqId, $originalExtension);
         $smartName = \sprintf('%s%s', $originalBasename, $uniqExtension);
+
         // Check if smartName is an acceptable size (some filesystems accept a max of 255)
         if (\strlen($smartName) <= 255) {
             return $smartName;
@@ -37,6 +38,7 @@ final class SmartUniqueNamer implements NamerInterface
             $shortBasename = \substr($originalBasename, 0, $shrinkBasenameSize);
 
             return \sprintf('%s%s', $shortBasename, $uniqExtension);
+
         // The extension is too long, but first try to preserve the basename, if possible
         // 253 is used to account for a dot and one letter extension
         } elseif (\strlen($uniqBasename) <= 253) {
@@ -44,14 +46,16 @@ final class SmartUniqueNamer implements NamerInterface
             // 254 is used to account for the dot
             $shrinkExtenstionSize = 254 - \strlen($uniqBasename);
             $shortExtension = \substr($originalExtension, 0, $shrinkExtensionSize);
-            
+
             return \sprintf('%s.%s', $uniqBasename, $shortExtension);
         }
+
         // Both the basename and extension are too long
         // 4 is used to keep three characters from the extension and include the dot
         $shrinkBasenameSize = 4 + \strlen($uniqId);
         $shortBasename = \substr($originalBasename, 0, $shrinkBasenameSize);
         $shortExtension = \substr($originalExtension, 0, 3);
+
         return \sprintf('%s.%s', $shortBasename, $shortExtension);
     }
 }
--- a/Naming/SmartUniqueNamer.php
+++ b/Naming/SmartUniqueNamer.php
@@ -57,6 +57,6 @@ final class SmartUniqueNamer implements NamerInterface
         $shortBasename = \substr($originalBasename, 0, $shrinkBasenameSize);
         $shortExtension = \substr($originalExtension, 0, 3);
 
-        return \sprintf('%s.%s', $shortBasename, $shortExtension);
+        return \sprintf('%s%s.%s', $shortBasename, $uniqId, $shortExtension);
     }
 }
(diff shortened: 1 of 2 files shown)


Naming/SmartUniqueNamer.php
```diff
@@ -17,10 +17,25 @@ public function name($object, PropertyMapping $mapping): string
     {
         $file = $mapping->getFile($object);
         $originalName = $file->getClientOriginalName();
-        $originalExtension = \pathinfo($originalName, PATHINFO_EXTENSION);
-        $originalBasename = \basename($originalName, '.'.$originalExtension);
+        $originalExtension = \strtolower(\pathinfo($originalName, PATHINFO_EXTENSION));
+        $originalBasename = \strtolower(\basename($originalName, '.'.$originalExtension));
```
garak: We can avoid strtolower here, it's already performed by Transliterator

Naming/SmartUniqueNamer.php
```diff
@@ -17,10 +17,25 @@ public function name($object, PropertyMapping $mapping): string
...
+        $uniqExtension = \sprintf('%s.%s', $uniqId, $originalExtension);
+        $smartName = \sprintf('%s%s', $originalBasename, $uniqExtension);
+        // Check if smartName is an acceptable size (some filesystems accept a max of 255)
+        if (\strlen($smartName) <= 255) {
+            return $smartName;
+        }
 
-        return \sprintf('%s%s.%s', $originalBasename, \str_replace('.', '', \uniqid('-', true)), $originalExtension);
+        // Shorten the basename to fit into 255 (excluding the unique extension)
+        $diffSize = (255 - \strlen($uniqExtension)) - \strlen($originalBasename);
```
garak: I understand that we don't want to cut out the extension, but I think that diff size should be calculated by taking extension itself into account

aewens: Fair enough, I assume there is no issue then using the `substr` function on the extension then to remove the excess?

garak: OK, let's take a little example: suppose that $smartName is 256 chars long, and extension is typical 3 letters + dot (like ".pdf"). We can deduct than $diffSize will be 252 (= 256 - 3 - 1) chars and so will fail check against 255.

aewens: This should be taken care of in commit [a26773c](https://github.com/aewens/VichUploaderBundle/commit/a26773ce1bf2533195b862fec9951b93b375d638#diff-43549d537b7d4f88517af8b366fd4a62).

aewens:
@garak The requested changes have been made, let me know if anything else needs to be done.

Naming/SmartUniqueNamer.php
```diff
@@ -17,10 +17,45 @@ public function name($object, PropertyMapping $mapping): string
     {
         $file = $mapping->getFile($object);
         $originalName = $file->getClientOriginalName();
-        $originalExtension = \pathinfo($originalName, PATHINFO_EXTENSION);
+        $originalExtension = \strtolower(\pathinfo($originalName, PATHINFO_EXTENSION));
         $originalBasename = \basename($originalName, '.'.$originalExtension);
         $originalBasename = Transliterator::transliterate($originalBasename);
+        $uniqId = \str_replace('.', '', \uniqid('-', true));
+        $uniqBasename = \sprintf('%s%s', $originalBasename, $uniqId);
+        $uniqExtension = \sprintf('%s.%s', $uniqId, $originalExtension);
```
garak: What's the point of concatenating uniqid to both basename **and** extension? 🤔

Naming/SmartUniqueNamer.php
```diff
@@ -17,10 +17,45 @@ public function name($object, PropertyMapping $mapping): string
...
+        if (\strlen($uniqExtension) <= 254) {
+            // Resize the basename to fit into 255 alongside the unique ID and extension
+            $shrinkBasenameSize = 255 - \strlen($uniqExtension);
+            $shortBasename = \substr($originalBasename, 0, $shrinkBasenameSize);
+
+            return \sprintf('%s%s', $shortBasename, $uniqExtension);
+
+        // The extension is too long, but first try to preserve the basename, if possible
+        // 253 is used to account for a dot and one letter extension
+        } elseif (\strlen($uniqBasename) <= 253) {
```
garak: No need for an "else" after a return

aewens: Right! Sorry about that, I'll have that changed in just a moment.

Naming/SmartUniqueNamer.php
aewens: On lines [44](https://github.com/aewens/VichUploaderBundle/blob/f0682b779f75e8da250699ac4e79cb0253f4c272/Naming/SmartUniqueNamer.php#L44), [47](https://github.com/aewens/VichUploaderBundle/blob/f0682b779f75e8da250699ac4e79cb0253f4c272/Naming/SmartUniqueNamer.php#L47), and [50](https://github.com/aewens/VichUploaderBundle/blob/f0682b779f75e8da250699ac4e79cb0253f4c272/Naming/SmartUniqueNamer.php#L50) the variable is used in the edge case where a file has an abnormally long extension; so the extension would be shortened instead of the basename. The variable helps reduce code duplication during that check, but I can move it down to that portion of the code to make it easier to see why it is being used.

Naming/SmartUniqueNamer.php
aewens: This has been fixed in commit [3cecc4c](https://github.com/aewens/VichUploaderBundle/commit/3cecc4c3e4c0ef831872a1cf097fc6be1ed605fa#diff-43549d537b7d4f88517af8b366fd4a62)

Naming/SmartUniqueNamer.php
aewens: This has been fixed in commit [3cecc4c](https://github.com/aewens/VichUploaderBundle/commit/3cecc4c3e4c0ef831872a1cf097fc6be1ed605fa#diff-43549d537b7d4f88517af8b366fd4a62)

aewens:
@garak The new tests have been added

PR #1055, opened 2019-11-10
Append docs regarding manually adding files

jhard:
Since it took me a while to figure out how to do this correctly. Since I was adding the same file multiple times which triggered an exception like `AbstractStorage.php line 40: SplFileInfo::getSize(): stat failed for misc/temp/5dc7fc9a66303/file.png`, I figured it might be a good idea to add a warning about this in there, too.

Edits more than welcome.


Resources/doc/known_issues.md
```diff
@@ -118,7 +118,16 @@ the image path to a new image in that instance of `File` and attempted a `flush(
 would happen, instead inject a new instance of `UploadedFile` with the new path to your new
 image to sucessfully trigger the upload.
 
-**N.B** : UploadedFile objects have a [*test* mode](http://api.symfony.com/3.4/Symfony/Component/HttpFoundation/File/UploadedFile.html#method___construct) that can be used to simulate file uploads.
+**N.B.** : UploadedFile objects have a [*test* mode](http://api.symfony.com/3.4/Symfony/Component/HttpFoundation/File/UploadedFile.html#method___construct) that can be used to simulate file uploads.
+If you are going to insert images programatically, make sure to set it to true, i.e.
+``` php
+<?php
+
+$uploadedFile = new \Symfony\Component\HttpFoundation\File\UploadedFile($filePath, basename($filePath), null, null, null, true);
```
garak: Signature of constructor changed in Symfony 4.1, so we need to add a distinction here

jhard: Good catch - I'm using 4.3 and the automatism in the constructor saved me, since I had looked at the constructor of the linked 3.4 documentation.

What's a sensable way to add that to docs? Have another short block saying "if you're using Symfony < 4.1, be aware that the constructor changed and you should use..."?

Should the link just above the proposed change be changed as well to point to the current version?

And btw, if you don't mind me asking, where did you look that up? Is there a smart list somewhere I don't know about, or is it just from memory because of working that much in the code?

garak:
Please rebase

jhard:
I hope I've done this correctly, I suck at git.

I've also changed the link to point to the current version 5.0, changed the code to match it, and added a note about the constructor's change.

garak:
Thank you @jhard

PR #1056, opened 2019-11-21
Allow use of Symfony 5

rachel-fizz:


garak:
Hi, thanks for your contribution.
It looks like supporting a wide range of Symfony versions (like 3.4-5.X) is too much for this bundle (and probably for other bundles too).
My idea about supporting Symfony 5 is to relase a new version of this bundle with support for only Symfony >=4.4, and continuing supporting Symfony 3.4 in previous versions.

rachel-fizz:
> Hi, thanks for your contribution.
> It looks like supporting a wide range of Symfony versions (like 3.4-5.X) is too much for this bundle (and probably for other bundles too).
> My idea about supporting Symfony 5 is to relase a new version of this bundle with support for only Symfony >=4.4, and continuing supporting Symfony 3.4 in previous versions.

Oh, ok. That would be fine too! Do you have any use for this PR then?

garak:
Well, I'm afraid that this PR is not very useful in its current form.
In understand that the idea I detailed above can be a bit harder for a newcomer, so maybe it's better to close this.

rachel-fizz:
Yep! I might be able to set up something like what you suggested, but unfortunately time is the main problem since I have to build applications for clients. Was kind of looking for a quick fix :)
Who knows, maybe I'll give it a shot if I have some spare time at work!

Issue #1067, opened 2019-12-04
Mapping location for autoconfiguration has changed


zghosts:
### BC Break Report

After upgrading to 1.12.0 vich expects the yaml or xml mapping file to be in:

`\dirname($ref->getFileName()).'/../config/vich_uploader';`
https://github.com/dustin10/VichUploaderBundle/blob/1.12.0/src/DependencyInjection/VichUploaderExtension.php#L94

changed from the following in 1.11.0
`$directory = \dirname($ref->getFileName()).'/Resources/config/vich_uploader';`
https://github.com/dustin10/VichUploaderBundle/blob/1.11.0/DependencyInjection/VichUploaderExtension.php#L94

Perhaps both locations should be checked

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 1.12.0

#### Summary
Documentation states that the mapping file should be in Resources/config/vich_uploader relative to the bundle class
however currently it expects it to be in ../config/vich_uploader

#### Previous behavior

a mapping file in Resources/config/vich_uploader will be automatically found

#### Current behavior

The mapping file will no longer be automatically found in its current location as its expected to be in ../config/vich_uploader relative to the bundle class

PR #1068, opened 2019-12-04
also check Resources/config/vich_uploader for mapping configuration

zghosts:
Fixes the BC break in as reported in issue #1067 by also checking Resources/config/vich_uploader if ../config/vich_uploader is not found


zghosts:
I'll update the readme.md and upgrade.md accordingly

garak:
I'm also thinking that it would be nice to add a deprecated message for the old path.
We'll remove it in 2.0

zghosts:
While that would be the way to go in order to remove support for it, as the documented directory layout for a bundle still recommends Resources/config I personally think you should hold off on removing support for it.

garak:
As already mentioned, documentation is just outdated.
Anyway, we still don't have a date for version 2 release, so it will be in the far future.

zghosts:
I'll add a deprecation notice aswell then

zghosts:
That means changing the layout of TestBundle in tests/Fixtures right?
Or actually creating a second test bundle with the new layout applied so both can be tested

garak:
Right, so let's drop the legacy test. Just change current TestBundle

zghosts:
I was reading up on the current state of the new layout deprecation.
quoting the changelog for 4.4/5.0
"As many bundles must be compatible with a range of Symfony versions, the current directory convention is not deprecated yet, but it will be in the future."

How about not triggering a deprecation until symfony deprecates the old layout.
But changing the test fixture to support the new layout.

garak:
This is mainly intended for project configuration.
Using a bundle (like AppBundle) for projects is deprecated since ages and I don't see a reason to avoid deprecation on our side

PR #1082, opened 2020-01-02
Add support for a ContainerInterface instance in FlysystemStorage

tgalopin:
I am currently working on integrating the official Flysystem bundle (https://github.com/thephpleague/flysystem-bundle) into this one and I realized there is a huge downside in the current implementation: the MountManager provided by OneUpFlysystemBundle is public and references *directly* all the filesystems defined with the bundle. This means that 1/ the container can't be optimized to remove unused filesystems from the compiled code and 2/ each time the MountManager service is instanciated, all the filesystem services are instanciated as well at the same time, even if they are not useful in the context. This hurts performance and can lead to unstable application if a storage tries to connect to a service which is down or inaccessible.

This PR adds support for a PSR ContainerInterface instance instead of a MountManager, to rely on the lazyness a container can have. It will also let me finalize the integration with the official bundle afterwards.

### DIFF
--- a/src/Storage/FlysystemStorage.php
+++ b/src/Storage/FlysystemStorage.php
@@ -5,25 +5,31 @@ namespace Vich\UploaderBundle\Storage;
 use League\Flysystem\FileNotFoundException;
 use League\Flysystem\FilesystemInterface;
 use League\Flysystem\MountManager;
+use Psr\Container\ContainerInterface;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
 use Vich\UploaderBundle\Mapping\PropertyMapping;
 use Vich\UploaderBundle\Mapping\PropertyMappingFactory;
 
 /**
  * @author Markus Bachmann <markus.bachmann@bachi.biz>
+ * @author Titouan Galopin <galopintitouan@gmail.com>
  */
 class FlysystemStorage extends AbstractStorage
 {
     /**
-     * @var MountManager Flysystem mount manager
+     * @var MountManager|ContainerInterface a registry to get FilesystemInterface instances
      */
-    protected $mountManager;
+    protected $registry;
 
-    public function __construct(PropertyMappingFactory $factory, MountManager $mountManager)
+    public function __construct(PropertyMappingFactory $factory, $registry)
     {
         parent::__construct($factory);
 
-        $this->mountManager = $mountManager;
+        if (!$registry instanceof MountManager && !$registry instanceof ContainerInterface) {
+            throw new \TypeError(sprintf('Argument 2 passed to %s::__construct() must be an instance of %s or an instance of %s, %s given.', __CLASS__, MountManager::class, ContainerInterface::class, is_object($registry) ? get_class($registry) : gettype($registry)));
+        }
+
+        $this->registry = $registry;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Storage/FlysystemStorage.php
```diff
@@ -5,25 +5,31 @@
...
 {
     /**
-     * @var MountManager Flysystem mount manager
+     * @var MountManager|ContainerInterface a registry to get FilesystemInterface instances
      */
-    protected $mountManager;
+    protected $registry;
 
-    public function __construct(PropertyMappingFactory $factory, MountManager $mountManager)
+    public function __construct(PropertyMappingFactory $factory, $registry)
```
garak: I think it would be better to keep the old parameter, making it optional, and add a third parameter. So, we can ensure BC.
Of course, this would require some logic to check if `MountManager` is passed directly or if it needs to be taken from registry

tests/Storage/Flysystem/MountManagerFlysystemStorageTest.php
```diff
@@ -0,0 +1,29 @@
...
+namespace Vich\UploaderBundle\Tests\Storage\Flysystem;
+
+use League\Flysystem\FilesystemInterface;
+use League\Flysystem\MountManager;
+
+/**
+ * @author Markus Bachmann <markus.bachmann@bachi.biz>
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ */
+class MountManagerFlysystemStorageTest extends AbstractFlysystemStorageTest
```
garak: This class needs to be final

tests/Storage/Flysystem/PsrContainerFlysystemStorageTest.php
```diff
@@ -0,0 +1,29 @@
...
+namespace Vich\UploaderBundle\Tests\Storage\Flysystem;
+
+use League\Flysystem\FilesystemInterface;
+use Psr\Container\ContainerInterface;
+
+/**
+ * @author Markus Bachmann <markus.bachmann@bachi.biz>
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ */
+class PsrContainerFlysystemStorageTest extends AbstractFlysystemStorageTest
```
garak: This class needs to be final

src/Storage/FlysystemStorage.php
tgalopin: AFAIK this is backward compatible in every ways and would allow us to deprecate the usage of MountManager if we want to in the future easily :) . I can change it to a third parameter of course, but I really think it's better suited here.

garak: You're right, it's already BC. But it forces to remove type hinting and doing manual checks on types.
I think that type hinting is more clear.
Feel free to add a deprecation here.

tgalopin: My problem about the alternative (3rd argument) is that it can lead to situations in which you have two arguments populated at the same time. It would not make sense.

I understand your concern about the typehinting. However I think this is more a limitation of the language than an actual architecture issue (whereas having a third argument would have architecture consequences). 

Here is my suggestion: instead of an InvalidArgumentException, I can throw a TypeError. This way, when PHP will support union typehints, the migration towards them will be seemless and backward compatible. WDYT?

garak: It seems a reasonable compromise.

tgalopin: And about the deprecation: OneUpFlysystemBundle should probably fix the issue first on their side, to avoid triggering deprecations without any ability to upgrade to a deprecation-free dependency IMO

tgalopin: I did the change :)

Issue #1091, opened 2020-01-17
MissingPackageException for "league/flysystem-bundle" on existing project w/ hand-wired "vich_uploader.storage.flysystem" service


mattjwhelan:
In one of my projects that uses this (fantastic) package I have configured the "vich_uploader.storage.flysystem" service manually and am passing into it a hand-wired Flysystem service, so I don't have either of the recommended Flysystem bundles installed.

![vich](https://user-images.githubusercontent.com/1261752/72645326-00f1ff00-3939-11ea-86e7-c9c6c2ccb4a8.png)

After successfully updating to 1.13.0 I am now encountering the new MissingPackageException at run time. Are the bundles absolutely necessary moving forward due to structural changes to Vich Uploader?

garak:
Signature was changed to accept an instance of container too, while before it was restricted to `MountManager`.
Anyway, I don't see a reason for such change to break your configuration... can you provide the stack trace of your exception?

mattjwhelan:
I don't have my app running on this computer, but I can get a stack trace later today.

At a glance I think what's breaking for me is in the new compiler pass here (line 32):

![missing](https://user-images.githubusercontent.com/1261752/72662935-6e8e4180-39b2-11ea-8e4e-868cc683fe26.png)

I don't have the "League\FlysystemBundle\FlysystemBundle" class installed in in my project as I'm not using the League bundle for my Flysystem integration (nor Oneup bundle).

garak:
We need @tgalopin here too

tgalopin:
Ah indeed, I didn't think about custom wiring of Flysystem. I'm working on a patch.

PR #1092, opened 2020-01-18
Fix Flysystem manual wiring BC break

tgalopin:
Fix https://github.com/dustin10/VichUploaderBundle/issues/1091

This allows back manual wiring of a Flysystem service in case no bundle is used.


tgalopin:
@mattjwhelan if you could ensure this fix your problem, that would be awesome :) !

mattjwhelan:
@tgalopin I can't run tests against my app at the moment, but I did review the changes in your branch and I think it looks good.

If no Flysystem bundles are installed then the compiler pass will not modify anything (or throw errors), which I think fixes the issue. Thanks!

PR #1093, opened 2020-01-18
Fix PHPUnit 8 warnings

tgalopin:
The following warnings were raised each time I ran the tests using PHPUnit 8:

```
There were 7 warnings:

1) Vich\UploaderBundle\Tests\Form\Type\VichFileTypeTest::testConfigureOptionsBC with data set #0 (array(true), array(true))
assertArraySubset() is deprecated and will be removed in PHPUnit 9.

2) Vich\UploaderBundle\Tests\Form\Type\VichFileTypeTest::testConfigureOptionsBC with data set #1 (array(false), array(false))
assertArraySubset() is deprecated and will be removed in PHPUnit 9.

3) Vich\UploaderBundle\Tests\Form\Type\VichFileTypeTest::testEmptyDownloadLinkDoNotThrowsDeprecation
assertArraySubset() is deprecated and will be removed in PHPUnit 9.

4) Vich\UploaderBundle\Tests\Form\Type\VichImageTypeTest::testConfigureOptionsBC with data set #0 (array(true), array(true))
assertArraySubset() is deprecated and will be removed in PHPUnit 9.

5) Vich\UploaderBundle\Tests\Form\Type\VichImageTypeTest::testConfigureOptionsBC with data set #1 (array(false), array(false))
assertArraySubset() is deprecated and will be removed in PHPUnit 9.

6) Vich\UploaderBundle\Tests\Form\Type\VichImageTypeTest::testEmptyDownloadLinkDoNotThrowsDeprecation
assertArraySubset() is deprecated and will be removed in PHPUnit 9.

7) Vich\UploaderBundle\Tests\Handler\UploadHandlerTest::testRemove
Method erase may not return value of type NULL, its return declaration is ": void"
```

This PR fixes these warnings.

### DIFF
--- a/tests/Handler/UploadHandlerTest.php
+++ b/tests/Handler/UploadHandlerTest.php
@@ -183,8 +183,7 @@ final class UploadHandlerTest extends TestCase
         $this->mapping
             ->expects($this->once())
             ->method('erase')
-            ->with($this->object)
-            ->willReturn(null);
+            ->with($this->object);
 
         $this->storage
             ->expects($this->once())
--- a/tests/Form/Type/VichFileTypeTest.php
+++ b/tests/Form/Type/VichFileTypeTest.php
@@ -40,7 +40,11 @@ class VichFileTypeTest extends TestCase
         $type->configureOptions($optionsResolver);
 
         $resolved = $optionsResolver->resolve($options);
-        $this->assertArraySubset($resolvedOptions, $resolved);
+
+        foreach ($resolvedOptions as $key => $value) {
+            $this->assertArrayHasKey($key, $resolved);
+            $this->assertSame($value, $resolved[$key]);
+        }
     }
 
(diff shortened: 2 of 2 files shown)


tests/Handler/UploadHandlerTest.php
```diff
@@ -183,8 +183,7 @@ public function testRemove(): void
         $this->mapping
             ->expects($this->once())
             ->method('erase')
-            ->with($this->object)
-            ->willReturn(null);
+            ->with($this->object);
```
garak: Why we don't check anymore the null returned?

tgalopin: Because of this:

```
7) Vich\UploaderBundle\Tests\Handler\UploadHandlerTest::testRemove
Method erase may not return value of type NULL, its return declaration is ": void"
```

PR #1095, opened 2020-01-20
Avoid the creation of a notnull constraint

CedCannes:
I don't know if it's the auto-validation added since symfony4.3 that does that. 
but if I don't add the ? in the annotation: 
@var File   ->  @var ?File
I have a notnull constraint that prevents the normal behaviour of the form even if it is configured with required=false
When I edit an uploadable without wanting to change the image, the validator blocks on "imageFile can't be null".

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -105,7 +105,7 @@ class Product
      * 
      * @Vich\UploadableField(mapping="product_image", fileNameProperty="imageName", size="imageSize")
      * 
-     * @var ?File
+     * @var File|null
      */
     private $imageFile;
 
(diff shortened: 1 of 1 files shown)


docs/usage.md
```diff
@@ -105,7 +105,7 @@ class Product
      * 
      * @Vich\UploadableField(mapping="product_image", fileNameProperty="imageName", size="imageSize")
      * 
-     * @var File
+     * @var ?File
```
garak: Can we use `@var File|null` instead?

CedCannes: Sure :)

I wanted to test the annotation proposed here https://github.com/symfony/symfony/pull/32107
 but it doesn't seem to be present in my symfony 5.0.2 project.

Issue #1097, opened 2020-01-30
DownloadHandler should set fallback filename in makeDisposition() call


TomaszGasior:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 1.13.0 

#### Summary

This is my controller action:

```
    /**
     * @Route("/file/{id}", name="file.download")
     */
    public function download(File $file, DownloadHandler $downloadHandler): Response
    {
        return $downloadHandler->downloadObject($file, 'fileData', null, true);
    }
```
I set 4th argument of downloadObject() to `true` to enforce downloading using original filename. Inside DownloadHandler, here https://github.com/dustin10/VichUploaderBundle/blob/master/src/Handler/DownloadHandler.php#L71 makeDisposition() method is called without setting fallback filename as third argument. Fallback filename must contain **only** ASCII characters. When fallback filename is not defined, makeDisposition() uses its second argument as both filename and fallback filename, see here https://github.com/symfony/http-foundation/blob/master/HeaderUtils.php#L169

#### Current behavior

When original filename contains non-ASCII characters, HeaderUtils::makeDisposition() fails with "The filename fallback must only contain ASCII characters." exception.

#### How to reproduce

To reproduce please create file with original filename containing non-ASCII characters, for example "Zażółć gęślą jaźń.docx" which contains Polish-specific characters.

#### Expected behavior

IMHO your bundle should define some basic/automatic fallback filename, date&time for example.

TomaszGasior:
@garak Could you please tell me when are you going to do new release with this fix?

garak:
Next week at last. Meanwhile, you can use our branch alias

TomaszGasior:
@garak Thanks.

PR #1098, opened 2020-01-30
DownloadHandler: set fallback filename in download response headers

TomaszGasior:
Fixes https://github.com/dustin10/VichUploaderBundle/issues/1097

### DIFF
--- a/src/Handler/DownloadHandler.php
+++ b/src/Handler/DownloadHandler.php
@@ -68,9 +68,12 @@ class DownloadHandler extends AbstractHandler
             \stream_copy_to_stream($stream, \fopen('php://output', 'wb'));
         });
 
+        $filename = \str_replace(['%', '/', '\\'], '', $filename);
+
         $disposition = $response->headers->makeDisposition(
             $forceDownload ? ResponseHeaderBag::DISPOSITION_ATTACHMENT : ResponseHeaderBag::DISPOSITION_INLINE,
-            $filename
+            $filename,
+            \filter_var($filename, FILTER_SANITIZE_STRING, FILTER_FLAG_STRIP_HIGH)
         );
         $response->headers->set('Content-Disposition', $disposition);
         $response->headers->set('Content-Type', $mimeType ?: 'application/octet-stream');
--- a/tests/Handler/DownloadHandlerTest.php
+++ b/tests/Handler/DownloadHandlerTest.php
@@ -46,16 +46,18 @@ class DownloadHandlerTest extends TestCase
     public function filenamesProvider(): array
     {
         return [
-            ['file_name', 'file_name'],
-            ['file_name.ext', 'file_name.ext'],
-            ['file-name.ext', 'file-name.ext'],
+            ['file name', 'file name', null],
+            ['file_name', 'file_name', null],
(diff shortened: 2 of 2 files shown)


src/Handler/DownloadHandler.php
```diff
@@ -70,7 +70,8 @@ private function createDownloadResponse($stream, string $filename, ?string $mime
 
         $disposition = $response->headers->makeDisposition(
             $forceDownload ? ResponseHeaderBag::DISPOSITION_ATTACHMENT : ResponseHeaderBag::DISPOSITION_INLINE,
-            $filename
+            $filename,
+            'download'.strrchr($filename, '.')
```
garak: This looks weak to me.
Can't we use something like `filter_var($filename, FILTER_SANITIZE_STRING, FILTER_FLAG_STRIP_HIGH)`?

TomaszGasior: Most browsers will use UTF-8 version it seems to me so I don't think we really need real file name as fallback. Ofc I can change this code but I don't know how FILTER_FLAG_STRIP_HIGH exactly works, I can't find exact documentation for this flag. From the other hand, here are HeaderUtils requirements: https://github.com/symfony/http-foundation/blob/master/HeaderUtils.php#L174 We can change hardcoded 'download` string to something better if we make sure alternative implementation won't cause another HeaderUtils exception.

garak: you can try on the string you provided in the issue. It just strips non ascii away

TomaszGasior: Yes, I tried. But still I am not sure what exactly does FILTER_FLAG_STRIP_HIGH so I can't check whether its output matches HeaderUtils requirements. Keep in mind that when user tries to download file, its much better to download "download.ext" file than to get 500 server error. :) Also, most users will never see fallback filename, most browsers support UTF-8 filenames. 

I will try to look at PHP source code when I will find more time, although.

garak: It just removes ascii codes higher than 127

TomaszGasior: @garak  Unfortunately FILTER_FLAG_STRIP_HIGH is not enough because /, \ and % are disallowed by HeaderUtils::makeDisposition(). Slashes are disallowed even in UTF-8 filename. I updated my pull request to filter it manually.

src/Handler/DownloadHandler.php
```diff
@@ -68,9 +68,12 @@ private function createDownloadResponse($stream, string $filename, ?string $mime
             \stream_copy_to_stream($stream, \fopen('php://output', 'wb'));
         });
 
+        $filename = str_replace(['%', '/', '\\'], '', $filename);
+
         $disposition = $response->headers->makeDisposition(
             $forceDownload ? ResponseHeaderBag::DISPOSITION_ATTACHMENT : ResponseHeaderBag::DISPOSITION_INLINE,
-            $filename
+            $filename,
+            filter_var($filename, FILTER_SANITIZE_STRING, FILTER_FLAG_STRIP_HIGH)
```
garak: this function must be called from root namespace (as well as any other functions)

garak:
Thank you @TomaszGasior

PR #1107, opened 2020-03-11
remove templating configuration

garak:
This is something we should have done when releasing version 1.13.0 (where templating support were removed).
So, this is now also removing configuration for templating and deprecating related option (we can't just remove it for the sake of BC)


alexsegura:
It broke my code 😭

```
The service "AppBundle\Controller\RestaurantController" has a dependency on  
   a non-existent service "vich_uploader.templating.helper.uploader_helper".
```

No worries, I will now rely on autowiring 🙂 
Maybe for 100% BC, add an alias for this service?

PR #1108, opened 2020-03-14
Allow setting custom id and class

### LINKED ISSUE #821: VichFileType widget does not allow setting custom id attribute in Twig
Usually, I can assign a custom id attribute for a form_widget within in a Twig as follows:
```
{{ form_widget(form.fieldName, {'id': 'myCustomId'}) }}
```
But this does not work, if the field to be rendered is a VichFileType.

I would like the VichFileType form type to be more consistent to Symfony's build-in form types. Allowing to assign a custom id would also be of great benefit, because it eases to access the field by JavaScript in order to do some validation.

Coffee2CodeNL:
Fixes #821

### DIFF
--- a/templates/Form/fields.html.twig
+++ b/templates/Form/fields.html.twig
@@ -5,7 +5,7 @@
 
 {%- block vich_file_widget -%}
     <div class="vich-file">
-        {{- form_widget(form.file) -}}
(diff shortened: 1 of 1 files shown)


garak:
Related issue was closed.
Thank you anyway

PR #1112, opened 2020-03-20
Make it possible to use 0 as a directory or filename.

Nexotap:
The testing if $name is empty has been made with empty(), which treats a string with value 0 as zero. 
In this case this seems not correct to me, since it should be able to name a directory or even a file as 0.

### DIFF
--- a/src/Naming/PropertyDirectoryNamer.php
+++ b/src/Naming/PropertyDirectoryNamer.php
@@ -72,7 +72,7 @@ class PropertyDirectoryNamer implements DirectoryNamerInterface, ConfigurableInt
         }
 
         // Check if string is empty (Better solution than empty())
-        if ($name === '') {
+        if ('' === $name) {
             throw new NameGenerationException(\sprintf('Directory name could not be generated: property %s is empty.', $this->propertyPath));
         }
 
--- a/src/Naming/PropertyDirectoryNamer.php
+++ b/src/Naming/PropertyDirectoryNamer.php
@@ -71,7 +71,8 @@ class PropertyDirectoryNamer implements DirectoryNamerInterface, ConfigurableInt
             throw new NameGenerationException(\sprintf('Directory name could not be generated: property %s does not exist.', $this->propertyPath), $e->getCode(), $e);
         }
 
-        if (empty($name)) {
+        // Check if string is empty (Better solution than empty())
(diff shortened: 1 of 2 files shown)


src/Naming/PropertyDirectoryNamer.php
```diff
@@ -71,7 +71,8 @@ public function directoryName($object, PropertyMapping $mapping): string
             throw new NameGenerationException(\sprintf('Directory name could not be generated: property %s does not exist.', $this->propertyPath), $e->getCode(), $e);
         }
 
-        if (empty($name)) {
+        // Check if string is empty (Better solution than empty())
+        if ($name === '') {
```
garak: We should also check if `$name` is null

src/Naming/PropertyNamer.php
```diff
@@ -68,7 +68,8 @@ public function name($object, PropertyMapping $mapping): string
             throw new NameGenerationException(\sprintf('File name could not be generated: property %s does not exist.', $this->propertyPath), $e->getCode(), $e);
         }
 
-        if (empty($name)) {
+        // Check if string is empty (Better solution than empty())
+        if ($name === '') {
```
garak: We should also check if `$name` is null

Nexotap:
Not sure what to do now because of the continuous-integration check. Do I have to recreate a PR?
I'm sorry, I'm really bad in this.

garak:
No need to redo PR, just add commits to this one.
Anyway, you should run tests _before_ pushing your code

Nexotap:
How do I run those tests?

Nexotap:
I got travis running now. But I have no clue what all those failed jobs are. Can you help me with them?

Nexotap:
I've read that, but I dont get it. 
I dont use docker, nor have I ever used travis. I barely even use github. Look, I'm just a guy who loves creating websites from time to time. While beeing happy, getting help from the community, whenever I can, I try giving something back to it.

garak:
I see.
Then I'll close this PR and I'll leave related issue open, waiting for someone else implementing a fix.
In the meantime, you can use your own directory namer as workaround. See [docs](https://github.com/dustin10/VichUploaderBundle/blob/master/docs/directory_namer/howto/create_a_custom_directory_namer.md)

PR #1120, opened 2020-04-21
Reduce number of calls to File::getMimeType() during upload.

le-yak:
I had to implement a custom [`MimeTypeGuesserInterface`](https://symfony.com/doc/current/components/mime.html#adding-a-mime-type-guesser), which is quite costly to run.

This simple fix removes some redundant calls to File::getMimeType().

_Context_
_I need to handle SVG images, for which PHP's MIME detection is [somewhat bogus](https://bugs.php.net/bug.php?id=79045)_
_I will gladly share my implementation of `MimeTypeGuesserInterface` on demand. I was also able to extract [dimensions](https://github.com/dustin10/VichUploaderBundle/blob/master/src/Storage/AbstractStorage.php#L44) from the `width` and `height` SVG attributes. I would submit another PR if you are interested._

### DIFF
--- a/src/Storage/AbstractStorage.php
+++ b/src/Storage/AbstractStorage.php
@@ -36,12 +36,13 @@ abstract class AbstractStorage implements StorageInterface
 
         $name = $mapping->getUploadName($obj);
         $mapping->setFileName($obj, $name);
+        $mimeType = $file->getMimeType();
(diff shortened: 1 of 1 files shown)


garak:
Thank you @le-yak

PR #1128, opened 2020-05-13
PropetyDirectoryNamer support for multiple properties

nuryagdym:
Hi,
I added some enhancements to support multiple properties.
The reason to this change I needed this kind of directory structure: 
```
directory_namer:
                service: vich_uploader.namer_directory_property
                options: { property: ['user.company.id', 'user.id']}
```
given this configuration (companyId = 20, userId = 10) it generates `20/10` directory


garak:
Bump? If not updated, I'll be forced to close without merging

PR #1131, opened 2020-05-19
Update namers.md

napestershine:
Fixed typo


napestershine:
I don't know why Tests are failing. But I only changed a word.

garak:
> I don't know why Tests are failing. But I only changed a word.

Don't worry, it's a known problem with an external lib.

PR #1132, opened 2020-06-02
Deprecation since Symfony 5.1: NodeDefinition::setDeprecated() requires 3 arguments

StudioMaX:
The signature of method `NodeDefinition::setDeprecated()` has been updated to `NodeDefinition::setDeprecation(string $package, string $version, string $message)`.

See https://github.com/symfony/symfony/blob/5.1/UPGRADE-5.1.md

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,7 @@ jobs:
         - php: 7.3
           env: SYMFONY_VERSION="^4.4.0" WITH_LIIP_IMAGINE=true
         - php: 7.4
-          env: VALIDATE_DOCS=false SYMFONY_VERSION="5.0.*"
-        - php: 7.4
-          env: VALIDATE_DOCS=false SYMFONY_VERSION="5.1.*"
+          env: VALIDATE_DOCS=false SYMFONY_VERSION="^5.1.0"
     allow_failures:
         - php: nightly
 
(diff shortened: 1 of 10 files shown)


composer.json
```diff
@@ -25,6 +25,7 @@
         "symfony/form": "^4.4 || ^5.0",
         "symfony/http-foundation": "^4.4 || ^5.0",
         "symfony/http-kernel": "^4.4 || ^5.0",
+        "symfony/mime": "^4.4 || ^5.0",
```
garak, line 28: Do we need mime?

StudioMaX: Yes, a few calls to `File::getMimeType` like this:
https://github.com/dustin10/VichUploaderBundle/blob/2903b5647621c9b5b5fbaac38cb481f50fc7d192/src/Storage/AbstractStorage.php#L39
And because of this (see https://github.com/symfony/symfony/blob/5.1/UPGRADE-5.1.md#httpfoundation): 
> HttpFoundation: Made the Mime component an optional dependency

.travis.yml
```diff
@@ -25,7 +25,9 @@ jobs:
         - php: 7.3
           env: SYMFONY_VERSION="^4.4.0" WITH_LIIP_IMAGINE=true
         - php: 7.4
-          env: VALIDATE_DOCS=false SYMFONY_VERSION="^5.0.0"
+          env: VALIDATE_DOCS=false SYMFONY_VERSION="5.0.*"
```
garak, line 28: No need to test every minor. Let's keep "^5.1.0" only

garak:
Thank you @StudioMaX

PR #1138, opened 2020-06-22
fix translations

### LINKED ISSUE #1137: vich_file_label create missing translations
### Bug Report

|    Q            |   A
|---------------- | ------
| BC Break        | no
| Bundle version  | 1.14.0 
| Symfony version | 4.4.9
| PHP version     | 7.4

#### Summary

Hi @garak, thanks for this useful bundle.

The way VichUploader translate the form label seems buggy to me and report missing translations.
```
{%- block vich_file_label -%}
    {%- set label = label|trans|default(label) -%}
    {{- block('form_label') -}}
{%- endblock -%}
```

Firstly, when using `'label' => false`, it use `trans` on `false` and report a missing translation for `''`. Same when providing no label, it use `trans` on `null`. This leads to an error with Symfony 5. (Cf https://github.com/dustin10/VichUploaderBundle/issues/1066)

Secondly, when using `'label' => 'myLabel'`, it use obviously `trans` on `myLabel`. But the block `form_label` try to translate the label too with the option `'translation_domain'`. This means that the profiler report a missing translation when trying the second translation.

Finally, when using  `'translation_domain' => false`, the label is still translated.

### Solutions

The difficulty would be to determine what is BC and what is BC-break

Currently, If I have the following translations
```
#messages.yaml
myLabel: foo
myLabel2: bar

#VichUploaderBundle.yaml 
myLabel3: baz
bar: hello
```

- `myLabel` will be translated to `foo` with the first trans. (and then foo will be a missing translation in VichUploaderBundle)
- `myLabel2` will be translated to `bar` with the first trans. And then it will be translated to `hello` by `form_label`.
- `myLabel3` will be reported as a missing translation and then will be translated to `baz`.

IMHO the best solution would be to
- add a `delete_translation_domain` option and use it instead of `translation_domain` for `delete_label`.
- add a `download_translation_domain` option and use it instead of `translation_domain` for `download_label`
- change the default value of `translation_domain` to `messages`.
- remove the `{%- set label = label|trans|default(label) -%}` code.
But this would need to release a new major version.

I didn't find a BC way to fix both the fact that `false/null` throw an error with Symfony 5 and the double translate calls which leads to missing translations.

garak:
Fix #1137

### DIFF
--- a/src/Form/Type/VichFileType.php
+++ b/src/Form/Type/VichFileType.php
@@ -65,11 +65,9 @@ class VichFileType extends AbstractType
             'asset_helper' => false,
             'download_link' => null,
             'download_uri' => true,
-            //TODO: use 'form.label.download'
-            'download_label' => 'download',
-            'delete_label' => 'form.label.delete',
+            'download_label' => 'vich_uploader.link.download',
+            'delete_label' => 'vich_uploader.form_label.delete_confirm',
             'error_bubbling' => false,
-            'translation_domain' => 'VichUploaderBundle',
         ]);
 
         $resolver->setAllowedTypes('allow_delete', 'bool');
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,10 @@
+Upgrading from v1.14.0 to 1.15.0
+================================
+
+- translation has been refactored to use standard domain. If you added translations in "VichUploaderBundle" domain, you can move it to "messages" domain.
+  Also, label were change to avoid possible conflicts: from "Download" to "vich_uploader.link.download", and from "form.label.delete" to
+  "vich_uploader.form_label.delete_confirm".
+
 Upgrading from v1.12.0 to 1.13.0
 ================================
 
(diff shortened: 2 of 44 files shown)


src/Form/Type/VichFileType.php
```diff
@@ -65,11 +65,10 @@ public function configureOptions(OptionsResolver $resolver): void
...
             'download_link' => null,
             'download_uri' => true,
-            //TODO: use 'form.label.download'
-            'download_label' => 'download',
-            'delete_label' => 'form.label.delete',
+            'download_label' => 'vich_uploader.link.download',
+            'delete_label' => 'vich_uploader.form_label.delete_confirm',
             'error_bubbling' => false,
-            'translation_domain' => 'VichUploaderBundle',
+            'translation_domain' => 'messages',
```
VincentLanglet, line 71: It's already the default value, I don't think you need this line.

VincentLanglet:
Hi @garak ! 
Do you plan to make a new release with this ? It'll avoid to require a specific commit or dev-master. :)

garak:
Yes, new version will be released between today and tomorrow

PR #1145, opened 2020-08-10
Added "delete_label" option to the documentation

Erdou:
Added the missing "delete_label" option to the docs.
Without this, "vich_uploader.form_label.delete_confirm" is shown (which also shows here an issue of translation)

### DIFF
--- a/docs/form/vich_file_type.md
+++ b/docs/form/vich_file_type.md
@@ -35,27 +35,13 @@ allow_delete
 
 delete_label
 --------------
-**type**: `string`, `callable`, `Symfony\Component\PropertyAccess\PropertyPath` **default**: `vich_uploader.form_label.delete_confirm`
-
-Can be string 
-```php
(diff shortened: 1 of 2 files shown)


docs/form/vich_file_type.md
```diff
@@ -32,6 +33,33 @@ allow_delete
...
+```php
+use Vich\UploaderBundle\Form\Type\VichImageType;
+
+$builder->add('genericFile', VichImageType::class, [
+    'delete_label' => 'delete_file',
+]);
+
+```
+
+Can be callable
```
garak, line 50: Are you sure label can be a callable?

Erdou: You're right, I quickly supposed that but contrary to _download_label_, it's not.
I'm correcting.

Erdou: Done (also removed Property path option as after testing it wasn't working neither, and improved the example to showcase an actual usage of it)

PR #1156, opened 2020-09-16
Add cancellability for pre events

### LINKED ISSUE #1155: Remove files async
|    Q            |   A
|---------------- | ------
| Bundle version  | 1.15.0
| Symfony version | 5.1.5
| PHP version     | 7.4.9

### Support Question

I currently use the bundle in combination with the `league/flysystem-bundle`. It works great, but I want to delete the files async, a solution would be to use the messenger which I already use, but if I use the `PRE_REMOVE` event, how can I stop the vich uploader from doing his job?

Could also be a Feature request ;)

chapterjason:
This pull request, adds the possibility to cancel events. This has the advantage to cancel the actual remove request and dispatch a message to make the file removal asynchronously.

Cancellation only works for the `pre_remove` event.

Fixes #1155

### DIFF
--- a/docs/events/events.md
+++ b/docs/events/events.md
@@ -15,6 +15,8 @@ The following is a list of events you can listen to:
 |`vich_uploader.pre_remove`|`Events::PRE_REMOVE`|before a file is removed|
 |`vich_uploader.post_remove`|`Events::POST_REMOVE`|after a file is removed|
 
+The `vich_uploader.pre_remove` event is cancelable, that means that the actual remove request will not take place, and you have to take action.
+
 Example
 -------
 
@@ -50,4 +52,4 @@ services:
             - { name: kernel.event_listener, event: vich_uploader.pre_upload }
 ```
 
-[Return to the index](index.md)
+[Return to the index](../index.md)
--- a/docs/events/howto/remove_files_asynchronously.md
+++ b/docs/events/howto/remove_files_asynchronously.md
@@ -1,0 +1,114 @@
+Remove files asynchronously
+=========================
+
+To remove files asynchronously, ensure you have the [Messenger](https://symfony.com/doc/current/messenger.html) installed.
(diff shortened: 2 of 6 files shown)


garak (changes requested):
We need to adapt documentation accordingly

chapterjason:
@garak I added some documentation, I also tested some more complex scenarios like sending the whole event or the mapping and object.

The issue there is that `UploadedFile` is not serializable, also to prevent any side effects, like not uploading the file, a user can clone the object and remote the uploading file from the object. But I think in any case it's not good to send to much in messages.

garak:
Do you think that make sense to allow event canceling outside for uploading and injecting?
Maybe we could add it for deleting only.

chapterjason:
@garak Yeah I thought about that and it doesn't make sense for me for upload and inject.

We could introduce a new `PreRemoveEvent` class for that event only, but that would be a breaking change.
To have the cancellation in the `Event` class and only use it for a single event might be confusing.

chapterjason:
@garak After thinking about that a bit, it could make sense to cancel the upload. Uploading to the local file storage first and serve the file from there and after a worker processed the file, moved to the desired file storage and updates the object it could be served from there. But I think that is a really rare scenario and would be too complex.

docs/events/howto/remove_files_asynchronous.md
```diff
@@ -0,0 +1,114 @@
+Remove files asynchronous
```
garak, line 1: I guess that "asynchronous" should be "asynchronously" here and in following occurrences.

garak (changes requested):

docs/events/howto/remove_files_asynchronous.md
chapterjason: Changed

chapterjason:
@garak I changed everything for the `pre_remove` event only.

garak (approved):

Issue #1177, opened 2021-02-04
Support for league/flysystem v2.0


WolfgangDrescher:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | probably yes

#### Summary

As `league/flysystem` and the corresponding symfony bundle `league/flysystem-bundle` released recently version 2.0. It would be nice if the VichUploaderBundle would support these new version as well. What I experienced so far is that the `League\Flysystem\FilesystemInterface` does no longer exist in `Vich\UploaderBundle\Storage\FlysystemStorage`.

mewebstudio:
I have the same problem.

digitaltim-de:
Same Issue here

fullbl:
How should it be done?

It's possible to create another storage (FlysystemV2Storage), but I see a set of problems:
- version 2 of FilesystemOperator (the old FilesystemInterface) doesn't allow to get the adapter, so don't know how to find the full path in doResolvePath method
- how can we test them? It's not possible to require two versions of the same package!
- is it possible to check which version of the bundle the user have to automatically load the right storage?

garak:
@fullbl the idea is to drop v1 and support only v2 from next minor version

digitaltim-de:
Any fix here in WIP?

PR #1178, opened 2021-02-08
Support for league/flysystem v2.0 

fullbl:
as in discussion #1177 I'm trying to create another storage, but, so far, I don't know how:

- get the full path in doResolvePath method, since getAdapter function has gone
- make tests work, since two versions of the same library should be needed


garak:
>     * make tests work, since two versions of the same library should be needed

Feel free to drop v1

fullbl:
this changes everything, I'll open another PR!

PR #1202, opened 2021-04-30
Make symfony/form a suggestion

### LINKED ISSUE #1201: Suggest symfony/form instead of requiring it
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | maybe

#### Summary

It would be good if this package did not have a hard dependency on `symfony/form`. A [suggestion](https://getcomposer.org/doc/04-schema.md#suggest) could be used instead.

Background: not every user of this package will use Symfony forms. The fact that there is an option to disable the Symfony form integration (`vich_uploader.form`) indicates that this was already kept in mind. An example without Symfony forms is [API platform](https://api-platform.com/docs/core/file-upload/).

Changing this could be a slightly breaking change if projects don't explicitly require `symfony/form`. Then again if projects really use Symfony forms then this is very unlikely. Also the integration option `vich_uploader.form` should then probably be disabled by default.

mbrodala:
Fixes #1201

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,8 @@
+Upgrading from v1.17.1 to ...
+================================
+
+- the "symfony/form" package is now a suggestion. Make sure to explicitly require this package if you use Symfony forms.
+
 Upgrading from v1.14.0 to 1.15.0
 ================================
 
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,6 @@
         "symfony/console": "^4.4 || ^5.0",
         "symfony/dependency-injection": "^4.4 || ^5.0",
         "symfony/event-dispatcher-contracts": "^1.1 || ^2.0",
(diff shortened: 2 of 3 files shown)


garak (changes requested):
I guess that if you remove a dependency, if you need to add it as dev-dependency then

mbrodala:
Yes, obviously. :facepalm: Fixed now.

garak (changes requested):
You can ignore failure on phpstan, it's a bug (not depending on us)
Anyway, we still need to:
1) document this change in UPGRADE.md
2) document the need for form component in types documentation

mbrodala:
Alright, did that now.

garak (approved):

garak:
Thank you @mbrodala !

PR #1212, opened 2021-06-16
Re-allow jms/metadata v1

guilliamxavier:
Refs https://github.com/dustin10/VichUploaderBundle/pull/960#discussion_r652571339 (updated as ^2.0 is now ^2.4)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     "require": {
         "php": "^7.3 || ^8.0",
         "ext-simplexml": "*",
-        "jms/metadata": "^2.4",
+        "jms/metadata": "^1.7 || ^2.4",
         "symfony/config": "^4.4 || ^5.0",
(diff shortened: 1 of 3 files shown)


guilliamxavier:
Looks like it needs a maintainer to trigger the CI (to check that it passes)

guilliamxavier:
Thanks for the build... which failed (https://github.com/dustin10/VichUploaderBundle/pull/1212/checks?check_run_id=2838775180).
So the problem is not on the *return* type declaration but on the *parameter* type declaration.
We could remove it (https://3v4l.org/3Gram); although that would be a potential BC break for possible overriders of `AbstractYamlDriver` / `XmlDriver` that would have added it, #960 was one already (by adding the return type declaration).
WDYT? (I'll try it as Draft anyway)

guilliamxavier:
Passing now 🙂

PR #1214, opened 2021-06-18
add final/internal annotations

garak:
This will allow to mark classes as "final" in next major version


andyexeter:
Hey,

What was the rationale for marking `Vich\UploaderBundle\VichUploaderBundle` as internal?

This causes PhpStorm (and maybe other IDEs?) to put a strikethrough in the class when loading it in `config/bundles.php`:

Screenshot from 2022-01-18 18-24-34

It's a minor issue, but it does make it look like the developer has done something wrong IMO :)

garak:
You're right, it should be removed.
Fancy opening a PR?

PR #1210, opened 2021-06-05
Allow PHP8 Attribute

mathieupetrini:
Add PHP 8 Attribute Support

PR #1219, opened 2021-08-16
Converted dimensions mapping type from array to string

yoshz:
Mapping dimensions should be a string containing the property name to map and not an array.

### DIFF
--- a/src/Mapping/Annotation/UploadableField.php
+++ b/src/Mapping/Annotation/UploadableField.php
@@ -46,7 +46,7 @@ class UploadableField implements AnnotationInterface
     protected $originalName;
 
     /**
-     * @var array
+     * @var string
      */
     protected $dimensions;
 
@@ -64,7 +64,7 @@ class UploadableField implements AnnotationInterface
         string $size = null,
         string $mimeType = null,
         string $originalName = null,
-        array $dimensions = null
+        string $dimensions = null
     ) {
         if (empty($options['mapping']) && empty($mapping)) {
             throw new \InvalidArgumentException('The "mapping" attribute of UploadableField is required.');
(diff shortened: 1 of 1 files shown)


garak:
Are you sure? What kind of string do you expect?

yoshz:
UploadableField only contains mapping to properties. So properties are always string. 
Not sure why the array type is added, it broke my configuration when I switched from annotations to attributes.
The dimensions value is indeed an array, but that is set on the model.

garak:
If dimensions is an array, the dimensions property should be an array of strings (a string for the first dimension, another string for the second dimension).

guilliamxavier:
I think @yoshz is right and @garak is probably confused (no offense). The attribute intended usage is e.g.:

```php
#[ORM\Entity]
#[Vich\Uploadable] 
class Product
{
    /* ... */

    #[Vich\UploadableField(/* ... */, dimensions: 'imageDimensions')]
    private ?File $imageFile = null;

    /* ... */

    #[ORM\Column(type: 'simple_array')]
    private ?array $imageDimensions = null;

    /* ... */
}
```

Product#imageDimensions is an array (of 2 ints), but UploadableField#dimensions is a string (property name).

This PR is a bugfix of the new feature introduced in #1210

garak:
> I think @yoshz is right and @garak is probably confused (no offense). The attribute intended usage is e.g.:

Hey, don't worry! There's no offence in pointing my mistakes.
Unfortunately, I don't use annotations nor new properties, and so I'm dubious about this proposal.
Maybe @mathieupetrini can confirm that this change is needed?

yoshz:
I know there is some confusion but this is a bug since #1210 is merged.
Can @mathieupetrini please confirm this bugfix?

kor3k:
yes this is a bug, if you use the attribute/annotation like this

```
#[Vich\UploadableField(..., dimensions: 'file.dimensions')]
```
then you get this error
```
Vich\UploaderBundle\Mapping\Annotation\UploadableField::__construct(): Argument #7 ($dimensions) must be of type ?array, string given
```

garak (approved):

PR #1220, opened 2021-08-19
Fix Collector toolbar spacing

IonBazan:
This tiny change fixes the padding and style in the debug toolbar.

Before:
image

After:
image

This is because profiler styles are explicitly looking for `<b>` there and not `<strong>`: https://github.com/symfony/web-profiler-bundle/blob/95fb24b09551688a09cffac95a2ddbb907833f07/Resources/views/Profiler/toolbar.css.twig#L177


garak:
Nice catch!

PR #1238, opened 2021-11-22
Remove return typehint

tacman:


garak:
I guess this one was just a mistake...

PR #1244, opened 2021-12-08
Rework Doctrine mapping to XML to fix issue with SF 5.4

tobias-93:
Different approach to #1243, change the configuration to XML (as suggested in https://symfony.com/doc/current/bundles/best_practices.html#doctrine-entities-documents).
Symlink in `src/Resources/config/doctrine` is needed for backwards compatibility, as the bundle path detection is fixed in Symfony 5.4.

### DIFF
--- a/config/doctrine/File.orm.xml
+++ b/config/doctrine/File.orm.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<!-- A copy of this file is necessary in src/Resources/config/doctrine for SF <5.4 compatibility. If this file is updated the copy needs to be updated too -->
+<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
+                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                  xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping
+                   https://raw.github.com/doctrine/doctrine2/master/doctrine-mapping.xsd">
+    <embeddable name="Vich\UploaderBundle\Entity\File">
+        <field name="name" nullable="true"/>
+        <field name="originalName" column="original_name" nullable="true"/>
+        <field name="mimeType" column="mime_type" nullable="true"/>
+        <field name="size" type="integer" nullable="true"/>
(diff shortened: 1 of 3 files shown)


tobias-93:
Apparently Doctrine doesn't do symlinks, so I made a copy of the mapping file in the directory Doctrine expects it

config/doctrine/File.orm.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
garak, line 1: What if we use a symlink here?

tobias-93: Unfortunately, Doctrine uses a `is_file` call here: https://github.com/doctrine/persistence/blob/2.2.x/lib/Doctrine/Persistence/Mapping/Driver/SymfonyFileLocator.php#L151 That means that a symlink will work in neither location, as `is_file` returns false for symlinks.

garak (approved):

src/Resources/config/doctrine/File.orm.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
RomanApunts, line 1: I needed to move this file to /uploader-bundle/Resources/config/doctrine/File.orm.xml to fix this error: 

```
In AbstractDoctrineExtension.php line 250:
  Specified non-existing directory "C:\SymfonyProjects\moyaplastApp\vendor\vich\uploader-bundle/Resources/config/doctrine" as Doctrine mapping source.
```

Maybe there are a problem?¿

ctissier: +1

Got the same issue here. It prevents the completion of my deployment. If I manually copy/move the file, it's ok, but on my production stack, I can't (and don't want to) manually change the files, everything being driven but git scripts.

There seems to be a mismatch between the two paths:

- `/uploader-bundle/Resources/config/doctrine/File.orm.xml` that the `AbstractDoctrineExtension.php` tries to reach 
- `/uploader-bundle/src/Resources/config/doctrine/File.orm.xml` where the file actually is

(note the `/src/` on the second path)

Anyone knows where to fix that so that the update to sf 5.4 doesn't break the bundle?

PR #1256, opened 2022-01-01
Add dimensions helper functions to File class

### LINKED ISSUE #1255: Add dimensions helper functions
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

I'd like to embedded `Vich\UploaderBundle\Entity\File` in a new project but there's something that holds me back from doing it.
I always like to add some helper functions for the `dimensions` field, to avoid to directly deal with raw array data, like this:

```php
    public function getWidth(): ?int
    {
        if ($this->getDimensions()) {
            return $this->getDimensions()[0];
        }

        return null;
    }

    public function getHeight(): ?int
    {
        if ($this->getDimensions()) {
            return $this->getDimensions()[1];
        }

        return null;
    }
    
    // This one is just a "plus" because I always put image dimensions on each <img> tag to avoid layout shifting
    public function getRenderedHtmlDimensions(): ?string
    {
        if ($this->getDimensions()) {
            return 'width="' . $this->getWidth() . '" height="' . $this->getHeight() . '"';
        }

        return null;
    }
```

Actually, if I want to implement those functions, I need to create my own `Embeddable` class and duplicate all the mapping inside it, and it annoy me a bit.

If you like the idea of those helpers functions, are you open to a PR?

Jupi007:
- Implement `getWidth()` and `getHeight()`, which respectively return image width and height ;
- Implement `getHtmlDimensions()` which returns HTML formatted dimensions.

Closes #1255

### DIFF
--- a/src/Entity/File.php
+++ b/src/Entity/File.php
@@ -78,4 +78,47 @@ class File
     {
         $this->dimensions = $dimensions;
     }
+
+    /**
+     * A simple shortcut to the image width.
+     * Similar to `$file->getDimensions()[0]`.
+     *
+     * @return int|null Returns `null` if dimensions array is itself null
+     */
+    public function getWidth(): ?int
+    {
+        return null !== $this->dimensions ? $this->dimensions[0] : null;
+    }
+
+    /**
+     * A simple shortcut to the image height.
+     * Similar to `$file->getDimensions()[1]`.
+     *
+     * @return int|null Returns `null` if dimensions array is itself null
+     */
+    public function getHeight(): ?int
+    {
+        return null !== $this->dimensions ? $this->dimensions[1] : null;
+    }
+
+    /**
+     * Format image dimensions for use with html (to avoid layout shifting).
+     *
+     * Usage in twig template:
+     * ```twig
+     * <img src="..." alt="..." {{ image.htmlDimensions }}>
+     * <!-- Will render: -->
+     * <img src="..." alt="..." width="..." height="...">
+     * ```
+     *
+     * @return string|null Returns `null` if dimensions array is itself null
+     */
+    public function getHtmlDimensions(): ?string
+    {
+        if (null !== $this->dimensions) {
+            return \sprintf('width=%s height=%s', $this->getWidth(), $this->getHeight());
+        }
+
+        return null;
+    }
 }


src/Entity/File.php
```diff
@@ -78,4 +78,47 @@ public function setDimensions(?array $dimensions): void
...
+    /**
+     * A simple shortcut to the image width.
+     * Similar to `$file->getDimensions()[0]`.
+     *
+     * @return int|null Returns `null` if dimensions array is itself null
+     */
+    public function getWidth(): ?int
+    {
+        return $this->getDimensions() ? $this->getDimensions()[0] : null;
+    }
```
garak, line 91: We can access directly the property. Moreover, we should avoi implicit casting to Boolean and check for null equality.

garak (changes requested):

src/Entity/File.php
Jupi007: You mean I need to do `null !== $this->getDimensions()`?

garak: No, I mean `null !== $this->dimensions`

Jupi007: Oh, yes, you're right. I'm stupid 🤦‍♂️
As I based this on my own code, I didn't noticed I was still use the getter, sorry.
I'll fix that at once :)

src/Entity/File.php
```diff
@@ -78,4 +78,47 @@ public function setDimensions(?array $dimensions): void
...
+    {
+        return null !== $this->dimensions ? $this->dimensions[1] : null;
+    }
+
+    /**
+     * Format image dimensions for use with html (to avoid layout shifting).
+     *
+     * Usage in twig template:
+     * ```twig
+     * <img src="..." alt="..." {{ image.getHtmlDimensions }}>
```
garak, line 109: The "get" here is superfluous

src/Entity/File.php
```diff
@@ -78,4 +78,47 @@ public function setDimensions(?array $dimensions): void
...
+     * <!-- Will render: -->
+     * <img src="..." alt="..." width="..." height="...">
+     * ```
+     *
+     * @return string|null Returns `null` if dimensions array is itself null
+     */
+    public function getHtmlDimensions(): ?string
+    {
+        if (null !== $this->dimensions) {
+            return 'width='.$this->getWidth().' height='.$this->getHeight();
```
garak, line 119: Maybe a sprintf can be better than a triple concatenation. Moreover, this approach is forcing to add a space that can be useless. Maybe it's better to return an initial space (and change comment accordingly)

garak (changes requested):

src/Entity/File.php
Jupi007: I agree for `sprintf`.

But is the useless space in rendered Html very important?
Imo, drop it in twig doesn't looks very good in terms of readability:

```twig
<img src="..." alt="..."{{ image.htmlDimensions }}>
```
And btw, this function will only be used on `img` tags. So it can returns null, but it shouldn't, as it's surly an image which should have dimensions.

src/Entity/File.php
Jupi007: Indeed, I didn't noticed the "get".

src/Entity/File.php
garak: let's keep the space, then

garak (approved):

src/Entity/File.php
```diff
@@ -78,4 +78,47 @@ public function setDimensions(?array $dimensions): void
...
+     * <!-- Will render: -->
+     * <img src="..." alt="..." width="..." height="...">
+     * ```
+     *
+     * @return string|null Returns `null` if dimensions array is itself null
+     */
+    public function getHtmlDimensions(): ?string
+    {
+        if (null !== $this->dimensions) {
+            return \sprintf('width=%s height=%s', $this->getWidth(), $this->getHeight());
```
guilliamxavier, line 119: Missing quotes around placeholders

Jupi007: Yes, indeed. In fact it was render a double quote.
But I now understand it was the twig html escape mechanism.
So this require to add the `|raw` filter.

I'll do a fix PR.

Issue #1252, opened 2021-12-23
delete_on_update not working


julien-lmnr:
### Bug Report

|    Q            |   A
|---------------- | ------
| BC Break        | no
| Bundle version  | dev-master
| Symfony version | 6.0.1
| PHP version     | 8.0.14

Hello everyone,

It seems that the **delete_on_update** parameter is not working correctly, at least not in my case with the **dev-master version**.
Strange because the **delete_on_remove** parameter works very well.

```php
#[Vich\Uploadable]
class PageImageBlock
{
    #[Vich\UploadableField(mapping: 'page_image', fileNameProperty: 'imageName')]
    private ?File $imageFile = null;

    #[ORM\Column(type: 'string', length: 255)]
    private ?string $imageName = null;

    public function getImageFile(): ?File
    {
        return $this->imageFile;
    }

    public function setImageFile(?File $imageFile): self
    {
        $this->imageFile = $imageFile;

        if (null !== $imageFile) {
            $this->updatedAt = new DateTime();
        }

        return $this;
    }

    public function getImageName(): ?string
    {
        return $this->imageName;
    }

    public function setImageName(?string $imageName): self
    {
        $this->imageName = $imageName;

        return $this;
    }
}
```

```yaml
vich_uploader:
    db_driver: orm

    metadata:
        type: attribute

    mappings:
        page_image:
            uri_prefix: '%app.path.page_image%'
            upload_destination: '%kernel.project_dir%/public%app.path.page_image%'
            namer: Vich\UploaderBundle\Naming\SmartUniqueNamer
```

Do you have an idea ?

garak:
Sorry, I can't reproduce the problem.
Using Symfony 6.0 and VichUploaderBundle 1.19, when I upload a new file I see that the old file is deleted.

Jupi007:
Hi,
it looks like I also have this problem.
Using _Sf 5.4_ and _VichUploaderBundle 1.19_

**delete_on_remove** works but not **delete_on_update**.

garak:
I was able to reproduce the problem with Symfony 5.4.1, but with Symfony 5.4.2 is working.

Jupi007:
Just tried with **Sf 5.4.2** (I was using the **5.4.1**).
I also removed `var/cache`, but it still doesn't work.

garak:
@Jupi007 I guess it's related to other dependencies. Try to upgrade all your vendors.
[...]

PR #1261, opened 2022-01-19
Remove internal annotation from bundle class

andyexeter:
As per the discussion in #1214 with @garak, the `@internal` annotation on `Vich\UploaderBundle\VichUploaderBundle` causes a strike-through to appear in the class name in `config/bundles.php` within PhpStorm and possibly other IDEs.

149996469-fe9a19fa-6097-47db-bb45-d42aae919c8f

This makes it look like the developer has done something wrong.

This PR simply removes the annotation from this class.


VincentLanglet:
@garak  Hi, would it be possible to create a new release 1.19.1 ?

garak:
> @garak Hi, would it be possible to create a new release 1.19.1 ?

1.19.1 will be released when we'll solve #1252

PR #1277, opened 2022-03-16
Fix mongodb changeset

IonBazan:
This PR tries to fix the entity/document listeners by:

- Relying on `Doctrine\Persistence` events
- Using `getObject()` from the event itself instead of using Adapter
~- Adding `getChangeSet()` to the adapter itself so it can be customized~

The MongoDB ODM part is a bit poorly tested as many classes are `final` which makes it really difficult to test.


IonBazan:
@garak I think this is now ready to go - just the PHPStan errors needs some fixing but most of it doesn't seem to be related to this PR itself.

IonBazan:
Unfortunately, this `AdapterInterface` change is mandatory:
- `getObjectFromArgs()` was removed because we can rely on `getObject()` on the event itself so the method becomes obsolete
- `getChangeSet()` was introduced to provide an adapter between MongoDB ODM (`getDocumentChangeSet()`) and ORM (`getEntityChangeSet()`) which is the core of the issue solved by this PR

If we really don't want to touch the interface, we could shift this logic to the listener itself but that would defeat the purpose of using the adapter, especially that the adapter is marked as `@internal` since v1.19.0.

I will take a look at PHPStan errors again and solve them in next commit 👍🏻

phpstan-baseline.neon
```diff
@@ -1,9 +1,9 @@
 parameters:
 	ignoreErrors:
 		-
-			message: "#^Method Vich\\\\UploaderBundle\\\\Adapter\\\\AdapterInterface\\:\\:recomputeChangeSet\\(\\) has no return type specified\\.$#"
+			message: "#^Call to an undefined method Doctrine\\\\Persistence\\\\ObjectManager\\:\\:getUnitOfWork\\(\\)\\.$#"
```
IonBazan, line 4: This error is ignored because the object returned from `getObjectManager()` is an instance of `Doctrine\ODM\PHPCR\DocumentManagerInterface` (https://github.com/doctrine/phpcr-odm/blob/master/lib/Doctrine/ODM/PHPCR/DocumentManagerInterface.php) but we are not installing this package so phpstan can't analyze this.

garak:
@IonBazan I'm not against the interface change, but this PR should include an update to changelog

garak:
Any news?

IonBazan:
@garak sorry, I think the issue that this PR was initially trying to fix has been solved in v1.19.1. I will rebase it and simplify the interface changes then.

IonBazan:
@garak I think this is now ready for a review again. Let me know if the changes still make sense after the changes from v1.19.
I've also added some `UPGRADING.md` notes to highlight the changes in `AdapterInterface`.

garak (approved):

garak:
Do you mind proposing the same changes on the master branch?

PR #1286, opened 2022-04-25
fix buildDeleteField

fabilocky:
change condition in buildDeleteField: no object and no uploaded file --> no delete button


garak:
This doesn't make sense to me. With your change, if the object is resolved to null, we don't return.

garak:
Moreover, if you find something that you consider to be a bug (or any other problem), please open an issue *before* opening a PR.
Thanks for understanding.

Issue #1273, opened 2022-02-24
Property path in form file type isn't respected


gglnx:
### Bug Report

The value `property_path` is ignored when using `VichFileType`. An `MappingNotFoundException` exception is thrown. 

|    Q            |   A
|---------------- | ------
| BC Break        | no
| Bundle version  | 1.19 
| Symfony version | 6.0.3
| PHP version     | 8.1.3

#### Summary

My `data_class` is using camel case (`myFile`) but for my form names I would like to use snake case (`my_file`). To add the mapping I configure every field with `property_path`.

#### Current behavior

`property_path` isn't used and the field tries to access the `data_class` by the form name, which doesn't exists and an `MappingNotFoundException` exception is thrown.

#### How to reproduce

```php
$builder->add('my_file', VichFileType::class, [
  'property_path' => 'myFile',
]);
```

The data class has a property `myFile`.

#### Expected behavior

The field should use the property path.

PR #1288, opened 2022-05-09
Allow to decorate the UploaderHelper

VincentLanglet:
Hi @garak. 

I recently tried the 1.19 version and got an issue.

Since the UploaderHelper is/will be final, I cannot extends it.
But since this class is required (and not an interface) in the construct of the UploaderExtension, I cannot use decoration either.

This PR solve this issue.

### DIFF
--- a/src/Templating/Helper/UploaderHelperInterface.php
+++ b/src/Templating/Helper/UploaderHelperInterface.php
@@ -8,9 +8,9 @@ interface UploaderHelperInterface
      * Gets the public path for the file associated with the
      * object.
      *
-     * @param object|array $obj       The object
-     * @param string|null  $fieldName The field name
-     * @param string|null  $className The object's class. Mandatory if $obj can't be used to determine it
+     * @param object      $obj       The object
+     * @param string|null $fieldName The field name
+     * @param string|null $className The object's class. Mandatory if $obj can't be used to determine it
      *
      * @return string|null The public asset path or null if file not stored
      */
--- a/src/Templating/Helper/UploaderHelperInterface.php
+++ b/src/Templating/Helper/UploaderHelperInterface.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Vich\UploaderBundle\Templating\Helper;
+
+interface UploaderHelperInterface
+{
+    /**
+     * Gets the public path for the file associated with the
+     * object.
+     *
+     * @param object|array $obj       The object
+     * @param string|null  $fieldName The field name
+     * @param string|null  $className The object's class. Mandatory if $obj can't be used to determine it
+     *
+     * @return string|null The public asset path or null if file not stored
+     */
+    public function asset($obj, ?string $fieldName = null, ?string $className = null);
+}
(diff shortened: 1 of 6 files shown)


VincentLanglet:
> The problem is that we would change the signature of the UploaderExtension constructor, and this can break if a user extended it.

Changing `UploaderHelper` to `UploaderHelperInterface` is not a BC-break since `UploaderHelper implements UploaderHelperInterface`.

Prior to this, if you pass UploaderHelper  to the constructor it will still works.
And if you pass `CustomUploaderHelper extends UploaderHelper` it will still works since it implements the `UploaderHelperInterface` interface thanks to UploaderHelper.

src/Templating/Helper/UploaderHelperInterface.php
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Vich\UploaderBundle\Templating\Helper;
+
+interface UploaderHelperInterface
+{
+    /**
+     * Gets the public path for the file associated with the
+     * object.
+     *
+     * @param object|array $obj       The object
```
garak, line 11: This looks weird to me, I see there's no consistence in the codebase, but anyway I think that we should only allow objects

garak:
> Changing `UploaderHelper` to `UploaderHelperInterface` is not a BC-break since `UploaderHelper implements UploaderHelperInterface`.

That's a good point.
So, I guess you only need to make tests passing.

src/Templating/Helper/UploaderHelperInterface.php
garak: @VincentLanglet what aout this?

VincentLanglet: Sorry, I forgot about this ; it's fixed

VincentLanglet: But then the third parameter is useless, isnt it ?

VincentLanglet:
@garak  I tried to add some deprecations. Tell me if it's ok for you.

src/Twig/Extension/UploaderExtension.php
```diff
@@ -41,6 +41,16 @@ public function getFunctions(): array
      */
     public function asset($object, ?string $fieldName = null, ?string $className = null): ?string
     {
+        if (!\is_object($object)) {
```
garak, line 44: This is already deprecated in the helper. If we keep it here too, user will get a duplicate deprecation warning, right?

VincentLanglet: Only if you don't use a custom UploaderHelper.

VincentLanglet: If you want to avoid duplicate deprecation, I can add a check `!$this->uploaderHelper instanceof UploadHelper`.
WDYT ?

VincentLanglet:
@garak I avoid the duplicate deprecations now

garak (approved):

garak:
Thank you @VincentLanglet

VincentLanglet:
> Thank you @VincentLanglet

You're welcome ; do you have time for a release ? :)

garak:
As soon as we sort up #1273

VincentLanglet:
Now https://github.com/dustin10/VichUploaderBundle/issues/1273 is solved, do you have time to create a new release @garak ?

garak:
Here you are https://github.com/dustin10/VichUploaderBundle/releases/tag/1.20.0

PR #1289, opened 2022-05-12
Fix misunderstandable documentation about using Attributes

welcoMattic:


garak (changes requested):
I appreciate the effort to make docs clearer, anyway I don't think that mentioning the path of configuration file could be better understood: the configuration can be located elsewhere and the only relevant part is the configuration path

welcoMattic:
As this file path is the one provided by the Flex recipe, I think it is the most relevant, don't you think?

garak:
> As this file path is the one provided by the Flex recipe, I think it is the most relevant, don't you think?

We need to stay abstract, we're not sure that everyone is using flex.
So, OK for mentioning the path, but only as a possibility

welcoMattic:
@garak fixed ;) thanks for your review

garak (approved):

PR #1290, opened 2022-05-28
Allow namer to be generic

VincentLanglet:


VincentLanglet:
Hi @garak,

this PR allows to write
```
/**
 * @phpstan-implements DirectoryNamerInterface<Incident>
 **/
class IncidentDirectoryNamer implements DirectoryNamerInterface
```
and then `phpstan` directly knows that the `directoryName` method only accept `Incident` object.

.github/workflows/build.yaml
```diff
@@ -12,6 +12,10 @@ jobs:
         steps:
             - name: Checkout
               uses: actions/checkout@v2
+            - name: Setup PHP
+              uses: shivammathur/setup-php@v2
+              with:
+                  php-version: 8.0
```
garak, line 18: Why we need this change? And why php 8.0?

garak (changes requested):

.github/workflows/build.yaml
VincentLanglet: You can look at the previous build https://github.com/dustin10/VichUploaderBundle/runs/6637173960?check_suite_focus=true

Since you're using a 8.0 docker https://github.com/OskarStark/phpstan-ga/blob/master/Dockerfile#L1 you have to install the dependency in 8.0

garak (approved):

PR #1292, opened 2022-06-24
Fix remove listener if property namer is used

1ed:
When the `PropertyDirectoryNamer` is used with the `id` of the entity
the remove listener throws an exception because it runs after the removal
and the id is no longer available at that time. This PR fixes the issue.

### DIFF
--- a/src/EventListener/Doctrine/RemoveListener.php
+++ b/src/EventListener/Doctrine/RemoveListener.php
@@ -43,7 +43,7 @@ class RemoveListener extends BaseListener
             if ($object instanceof Proxy) {
                 $object->__load();
             }
-            $this->entities[] = $object;
+            $this->entities[] = clone $object;
         }
     }
 
@@ -54,5 +54,6 @@ class RemoveListener extends BaseListener
                 $this->handler->remove($object, $field);
             }
         }
+        $this->entities = [];
     }
 }


src/EventListener/Doctrine/RemoveListener.php
```diff
@@ -54,5 +54,6 @@ public function postFlush(): void
                 $this->handler->remove($object, $field);
             }
         }
+        $this->entities = [];
```
garak, line 57: What's the purpose of this line?

1ed: It resets the internal state. It removes the references to the entities and frees memory, and if the listener runs multiple times the handler will not be called again for the same entities.

garak: Can you provide an example of a case where the listener runs multiple times?

1ed: Of course, for example if the entity manager flushed multiple times, e.g. in a batch process or at context/transaction boundaries (and apart from that I think making services as stateless as possible is a best practice).

```php
$photos = $entityManager->getRepository(FeaturedPhoto::class)->findAll();

$entityManager->remove(array_shift($photos));
$entityManager->flush();

$entityManager->remove(array_shift($photos));
$entityManager->flush();
```

image

1ed:
Thanks!

PR #1305, opened 2022-07-21
Implement getAllClassNames in AnnotationDriver

### LINKED ISSUE #1232: Add command autocompletion
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

As suggested in the [Symfony blog](https://symfony.com/blog/new-in-symfony-5-4-console-autocompletion), we can implement autocompletion for two of our commands: `vich:mapping:debug` and `vich:mapping:debug-class`

endelwar:
Make getAllClassNames in AnnotationDriver return a list of Doctrine managed entities that are annotated with `@Vich\Uploadable` by injecting `Doctrine\Persistence\ManagerRegistry` (Service ID `doctrine` in a Symfony app), making `vich:mapping:list-classes` discover that entities.

Also provides autocompletion for `vich:mapping:debug-class` (fix #1232)

[asciicast](https://asciinema.org/a/510200)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,7 @@
     "require": {
         "php": "^7.3 || ^8.0",
         "ext-simplexml": "*",
+        "doctrine/persistence": "^1.3 || ^2 || ^3",
         "jms/metadata": "^1.7 || ^2.4",
         "symfony/config": "^4.4 || ^5.0 || ^6.0",
         "symfony/console": "^4.4 || ^5.0 || ^6.0",
--- a/config/mapping.xml
+++ b/config/mapping.xml
@@ -16,7 +16,8 @@
 
         <!-- drivers -->
         <service id="vich_uploader.metadata_driver.annotation" class="Vich\UploaderBundle\Metadata\Driver\AnnotationDriver" public="false">
-            <argument type="service" id="vich_uploader.metadata.reader" />
+            <argument key="$reader" type="service" id="vich_uploader.metadata.reader" />
+            <argument key="$managerRegistry" type="service" id="doctrine" />
         </service>
 
         <service id="vich_uploader.metadata_driver.xml" class="Vich\UploaderBundle\Metadata\Driver\XmlDriver" public="false">
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -24,7 +24,4 @@ Searches for uploadable classes.
 php bin/console vich:mapping:list-classes
 ```
 
-> **NOTE** Only classes configured using XML or YAML are displayed.
-
-
 [Return to the index](index.md)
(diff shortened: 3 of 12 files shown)


endelwar:
I need help for making `VichUploaderBundleTest` pass: how can I make service `doctrine` (or service `Doctrine\Persistence\ManagerRegistry`) available to test kernel?

garak:
> I need help for making `VichUploaderBundleTest` pass: how can I make service `doctrine` (or service `Doctrine\Persistence\ManagerRegistry`) available to test kernel?

I'm not really familiar with that test, I tried to look into it and it looks like it's a functional one. So, the kernels used are instantiated directly and I guess that you can just register other bundles, like we already do with SimpleAppKernel.

garak (approved):

IonBazan:
@endelwar I think this PR breaks stuff if DoctrineBundle is not installed (for example, when using Doctrine MongoDB ODM or PHPCR instead). Any chance this could be improved?

```
The service "vich_uploader.metadata_driver.annotation" has a dependency on a non-existent service "doctrine"
```

endelwar:
The implementation is based on `Doctrine\Persistence\ManagerRegistry` which is an library (`doctrine/persistence`) required by:
- [doctrine/doctrine-bundle](https://github.com/doctrine/DoctrineBundle/blob/57280cea7b1a9212fde52a3b5065b8196ed213ce/composer.json#L36)
- [doctrine/mongodb-odm-bundle](https://github.com/doctrine/DoctrineMongoDBBundle/blob/f5ca2833d08e4107b3822e7da2b784aa68e8c4cb/composer.json#L19)
- Doctrine PHPCR (via [symfony/doctrine-bridge](https://github.com/symfony/doctrine-bridge/blob/68b53b14f337dbc6f92cf6f1656a0adad42482e0/composer.json#L21))

All the above libraries have a Registry based on `Doctrine\Persistence\AbstractManagerRegistry`, Symfony should map automatically this as `doctrine` service ID.

I've never used PHPCR, and only played aroud with doctrine/mongodb-odm-bundle, can you help me setting up a demo Symfony app for ODM and PHPCR, to further investigate the issue?

IonBazan:
Thanks for explaining @endelwar, noted!
While basing on `doctrine/persistence` is good, relying on `doctrine` service won't work as it is only available in ORM Bundle - MongoDB ODM and PHPCR call it `doctrine_mongodb` and `doctrine_phpcr` accordingly, not to mention that users may have all 3 installed in their apps. I will try to create a demo app to test this issue but we would most likely need to create a separate service to list all classes for each manager for each registry.

PR #1306, opened 2022-07-24
improve support for php 8

garak:
* improving use on PHP8
* improving docs to be more PHP8-friendly
* making docs more consistent with default recipe
* add a testing image for 8.1

### DIFF
--- a/docs/configuration_reference.md
+++ b/docs/configuration_reference.md
@@ -17,7 +17,7 @@ vich_uploader:
         cache: file
         type: annotation    # or attribute
     mappings:
-        product_image:
+        products:
             uri_prefix: /uploads    # uri prefix to resource
             upload_destination: ~   # gaufrette storage fs id, required
             namer: ~                # specify a file namer service for this entity, null default
--- a/docs/known_issues.md
+++ b/docs/known_issues.md
@@ -10,6 +10,9 @@ is the only updated.
 A workaround to solve this issue is to manually generate a change:
 
 ```php
+<?php
+
+use Doctrine\ORM\Mapping as ORM;
 use Symfony\Component\HttpFoundation\File\File;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
 
@@ -17,12 +20,8 @@ class Product
 {
     // ...
 
-    /**
-     * @ORM\Column(type="datetime")
-     *
-     * @var \DateTimeInterface|null
-     */
-    private $updatedAt;
+    #[ORM\Column(type: "datetime")]
+    private ?\DateTimeInterface $updatedAt = null;
 
     // ...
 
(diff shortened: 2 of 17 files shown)


garak:
What do you think @endelwar @VincentLanglet ?

docs/configuration_reference.md
```diff
@@ -15,9 +15,9 @@ vich_uploader:
     metadata:
         auto_detection: true
         cache: file
-        type: annotation    # or attribute
+        type: attribute    # or annotation, on PHP 7
```
VincentLanglet, line 18: It can still be annotation if annotations are used on PHP8, so it's not really related to the PHP version.

docs/known_issues.md
```diff
@@ -83,7 +83,7 @@ use Vich\UploaderBundle\Mapping\Annotation as Vich;
 class Product
 {
     /**
-     * @Vich\UploadableField(mapping="product_image", fileNameProperty="imageName")
+     * @Vich\UploadableField(mapping="products", fileNameProperty="imageName")
```
VincentLanglet, line 86: Here you're using annotation.

If you want to be consistent with the config, you should use attribute everywhere

docs/usage.md
```diff
@@ -261,81 +260,45 @@ class Product
...
     /**
      * If manually uploading a file (i.e. not using Symfony Form) ensure an instance
-     * of 'UploadedFile' is injected into this setter to trigger the  update. If this
+     * of 'UploadedFile' is injected into this setter to trigger the update. If this
      * bundle's configuration parameter 'inject_on_load' is set to 'true' this setter
      * must be able to accept an instance of 'File' as the bundle will inject one here
      * during Doctrine hydration.
      *
-     * @param File|UploadedFile|null $imageFile
+     * @param File|\Symfony\Component\HttpFoundation\File\UploadedFile|null $imageFile
```
VincentLanglet, line 299: Why not the import anymore ?

src/DependencyInjection/Configuration.php
```diff
@@ -78,7 +78,7 @@ private function addMetadataSection(ArrayNodeDefinition $node): void
                     ->fixXmlConfig('directory', 'directories')
                     ->children()
                         ->scalarNode('cache')->defaultValue('file')->end()
-                        ->scalarNode('type')->defaultValue('annotation')->end()
+                        ->scalarNode('type')->defaultValue(PHP_MAJOR_VERSION > 7 ? 'attribute' : 'annotation')->end()
```
VincentLanglet, line 81: This is a BC break for php8 user which didn't migrated from annotation.
(I am one of them).

I do appreciate the idea, but it require a major release...

docs/configuration_reference.md
garak: The idea is to encourage the use of attributes, so having a default of "attribute" in recent PHP versions. You can still use "annotation" configuration with PHP8, but you need to go explicit

src/DependencyInjection/Configuration.php
garak: It is indeed. It's noted in the CHANGELOG

docs/usage.md
garak: Just to be consistent with other examples

src/DependencyInjection/Configuration.php
VincentLanglet: People might miss the changelog when bumping dependencies if their composer use the ^ constraint. That's why it require a new major version Imho

VincentLanglet (approved):

endelwar:
LGTM 👍 
Like Vincent said, modification in `src/DependencyInjection/Configuration.php` should be made in a new major release.

PR #1307, opened 2022-07-24
Improve output of vich:mapping:debug-class

### LINKED ISSUE #771: MappingDebugClassCommand missing info
The command is currently displaying following info:
* mapping
* field name
* filename name

if configured, it should also diplay:
* size name
* mimeType name
* originalName name

endelwar:
Fix #771

Schermata 2022-07-25 alle 10 49 10


endelwar:
I've added the new information also to WebProfiler.
Schermata 2022-07-25 alle 11 37 19

garak (approved):

Issue #1308, opened 2022-07-28
Dimensions handler incorrect array handling when retrieving width


krstns:
### BC Break Report

After update to the latest version, on entities with embedded Uplaodable, if the file is not assigned, the new dimensions handler is crashing.

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 1.20

#### Summary

When retrieving entities without files (the property is null) the `getWidth` function throws an error.

#### Previous behavior

Entity is retrieved without issue

#### Current behavior

```
"detail": "Warning: Undefined array key 0",
	"trace": [
		{
			"namespace": "",
			"short_class": "",
			"class": "",
			"type": "",
			"function": "",
			"file": "\/srv\/api\/vendor\/vich\/uploader-bundle\/src\/Entity\/File.php",
			"line": 90,
			"args": []
		}
```

#### How to reproduce

Here's the entity (excerpt) in question:
```
<?php
namespace App\Entity\Direct;
use ApiPlatform\Core\Annotation\ApiResource;
use Vich\UploaderBundle\Entity\File as EmbeddedFile;
use DateTime;
use Doctrine\Common\Collections\ArrayCollection;
use Doctrine\Common\Collections\Collection;
use Doctrine\ORM\Mapping as ORM;
use Symfony\Component\HttpFoundation\File\File;
use Symfony\Component\HttpFoundation\File\UploadedFile;
use Symfony\Component\Uid\Uuid;
use Vich\UploaderBundle\Mapping\Annotation as Vich;
/**
 * A direct client
 */
#[ApiResource(
    collectionOperations: [
        "get",
        "post" => [
            "access_control" => "is_granted('ROLE_ADMIN')"
        ]
    ],
    itemOperations: [
        "get"
    ],
    shortName: "clinic",
    routePrefix: "/veterinaryinternal/1"
)]
#[Vich\Uploadable]
#[ORM\Entity]
#[ORM\Table(name: direct_client')]
class DirectClient {
    // initiated with system's current api version, used when creating new clinics
    private int $currentApiVersion = 2;
    /**
     * @var Uuid|null The id of this Veterinary Clinic.
     */
    #[ORM\Id]
    #[ORM\GeneratedValue(strategy: 'CUSTOM')]
    #[ORM\CustomIdGenerator(class: "doctrine.uuid_generator")]
    #[ORM\Column(name: 'dc_id', type: 'uuid', unique: true)]
    private ?Uuid $id;
[...]

PR #1309, opened 2022-07-28
#1308 Handle empty dimensions array case

krstns:
When normalizing the File the `getWidth` and `getHeight` are called and while the dimension for null files is returned as an empty array the original check is not enough to handle this case.

### DIFF
--- a/src/Entity/File.php
+++ b/src/Entity/File.php
@@ -87,7 +87,7 @@ class File
      */
     public function getWidth(): ?int
     {
-        return null !== $this->dimensions ? $this->dimensions[0] : null;
+        return isset($this->dimensions[0]) ? $this->dimensions[0] : null;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Entity/File.php
```diff
@@ -87,7 +87,7 @@ public function setDimensions(?array $dimensions): void
      */
     public function getWidth(): ?int
     {
-        return null !== $this->dimensions ? $this->dimensions[0] : null;
+        return isset($this->dimensions[0]) ? $this->dimensions[0] : null;
```
garak, line 90: Please let's simplify this: `return $this->dimensions[0] ?? null`

src/Entity/File.php
```diff
@@ -98,7 +98,7 @@ public function getWidth(): ?int
      */
     public function getHeight(): ?int
     {
-        return null !== $this->dimensions ? $this->dimensions[1] : null;
+        return isset($this->dimensions[1]) ? $this->dimensions[1] : null;
```
garak, line 101: same as above

garak (changes requested):

src/Entity/File.php
krstns: *facepalm* obviously...

krstns:
Requested simplification done.

garak (approved):

PR #1313, opened 2022-08-11
Make AnnotationDriver compatible with all supported adapters

### LINKED ISSUE #1312: Relying on Doctrine ORM Bundle
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 1.21.0

#### Summary

This bundle relies on `doctrine` service when Doctrine ORM Bundle is not installed (eg. Doctrine MongoDB ODM is used instead).

Introduced in #1305 (see discussion there for reference)

#### Previous behavior

This bundle did not require Doctrine ORM Bundle to be installed.

#### Current behavior

Following error is thrown if Doctrine Bundle is not installed:
```
The service "vich_uploader.metadata_driver.annotation" has a dependency on a non-existent service "doctrine"
```

#### How to reproduce

Upgrade from v1.20.0 to v1.21.0 and run `bin/console c:c`.

endelwar:
Fix #1312

### DIFF
--- a/config/mapping.xml
+++ b/config/mapping.xml
@@ -17,7 +17,7 @@
         <!-- drivers -->
         <service id="vich_uploader.metadata_driver.annotation" class="Vich\UploaderBundle\Metadata\Driver\AnnotationDriver" public="false">
             <argument key="$reader" type="service" id="vich_uploader.metadata.reader" />
-            <argument key="$managerRegistry" type="service" id="doctrine" />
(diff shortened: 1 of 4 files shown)


endelwar:
Ready to review @garak @IonBazan

IonBazan (approved):
LGTM! Thanks for fixing it so fast 🚀

garak (approved):

PR #1314, opened 2022-08-13
version 2

garak:


garak:
This is ready to be merged.
I'd like to get some feedback from recent active contributors:  @endelwar  @UlrichThomasGabor @krstns @VincentLanglet @nuryagdym @StudioMaX

src/Mapping/Annotation/UploadableField.php
```diff
@@ -13,60 +13,20 @@
...
     public function __construct(
-        string $mapping,
-        string $fileNameProperty = null,
-        string $size = null,
-        string $mimeType = null,
-        string $originalName = null,
-        string $dimensions = null
+        private readonly string $mapping,
+        private readonly ?string $fileNameProperty = null,
+        private readonly ?string $size = null, private readonly ?string $mimeType = null,
```
UlrichThomasGabor, line 26: I think there should be a newline

src/Form/Type/VichFileType.php
```diff
@@ -26,35 +25,14 @@
...
-    protected $propertyAccessor;
+    protected PropertyAccessorInterface $propertyAccessor;
 
     public function __construct(
-        StorageInterface $storage,
-        UploadHandler $handler,
-        PropertyMappingFactory $factory,
+        protected StorageInterface $storage,
+        protected UploadHandler $handler,
+        protected PropertyMappingFactory $factory,
```
UlrichThomasGabor, line 33: I would guess most of the constructor-injected services across all files could also be marked as `readonly`?

garak: You're right. I hope to have caught every occurence

UlrichThomasGabor:
I personally do not see the benefit of marking (so many) classes as `final`.

I've encountered the keyword a few times in different third-party projects recently and it was basically always in my way to easily reuse existing code when modifying the third-party projects behavior, which then either forces me to copy&paste code, which is obviously bad, or start a PR to get rid of the keyword, which is often cumbersome.

On the other hand, in my twenty years of software developing I have never though "Good that this class is final, I nearly accidentally would have extended it".

[I've also read the first good-looking Blog post on that topic now](https://ocramius.github.io/blog/when-to-declare-classes-final/). While some arguments are good, some also directly speak against usages here, compare the quoted guideline:
```
Final classes only work effectively under following assumptions:

1. There is an abstraction (interface) that the final class implements
2. All of the public API of the final class is part of that interface
```
and then this occurrence:
https://github.com/dustin10/VichUploaderBundle/blob/2b80a1bc6c10785dfa5fc49ad06f3012cd049679/src/Metadata/MetadataReader.php
which does not implement an interface.

I would not consider myself the master of software architecture, and I'm certainly more the fan of practical than theoretically-sound solutions, but from my point of view `final` is adding little to no practical benefit, but effectively prevents the re-use of existing code. Basically, every line of code in a `final` class is not reusable.

For example:
https://github.com/dustin10/VichUploaderBundle/blob/2b80a1bc6c10785dfa5fc49ad06f3012cd049679/src/Storage/FlysystemStorage.php
I cannot implement an extended FlysystemStorage based off that one.
- I cannot extend the class.
- I cannot composite an object of that class in a new adapter because the `getFilesystem` method is protected.

Whatever I want to do with a FlysystemAdapter replacement, I cannot re-use any of the existing code for it. This may not be intended, this may also be changed when I open a PR, e.g. by making the `getFilesystem` method public, but the as-is state is that the whole code of that file is a dead end. I am not particularly fond of that.

UlrichThomasGabor:
Besides that, I read over everything. Mostly it's only cleanup. I can check later if anything breaks in the project I integrated it lately, but my first guess is that the PR is fine.

VincentLanglet:
> I personally do not see the benefit of marking (so many) classes as `final`.

For maintaining another library, I can say that having final classes is so more simple to maintain.

There was `@final` for some time, already. People could complain about it.
For instance I created https://github.com/dustin10/VichUploaderBundle/pull/1288

If people encounter issue during the migration from 1 to 2, then can still add missing Interfaces/change method visibility if needed.

garak:
We mostly want to reduce the cost of maintenance to the possible lowest, plus allow us to change as much as possible without breaking public services.
By the way, the FlysystemAdapter is respecting the aforementioned assumptions.

UlrichThomasGabor:
May I ask which concrete maintenance task is becoming so much easier by adding the `final` keyword?

Changes to interfaces do not require a major version bump?

garak:
> May I ask which concrete maintenance task is becoming so much easier by adding the `final` keyword?

Of course: since you cannot extend the class, we are free to change the non-pubilc methods of the class in minor releases.

VincentLanglet:
> Of course: since you cannot extend the class, we are free to change the non-pubilc methods of the class in minor releases.

There are also some changes which can be considered as BC break for public method but are not anymore when the class is final.

For instance, changing
```
public function get(Foo $foo): object
{
    // ...
}
```
to
```
public function get(Foo|Bar $foo): Foo
{
    // ...
}
```
or 
```
public function get(FooInterface $foo): FooInterface
{
    // ...
}
```

nuryagdym:
Hi,
I will probably able to check the changes on this weekend.
But for now, shouldn't PHP version requirement in `composer.json`  be `^7.3 || ^8.1`. 
Library is using `readonly` properties which is available starting with PHP 8.1

garak:
@nuryagdym related change is here https://github.com/dustin10/VichUploaderBundle/pull/1314/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R18

nuryagdym:
> @nuryagdym related change is here https://github.com/dustin10/VichUploaderBundle/pull/1314/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R18

sorry, my bad, when I wrote it **Files Changed** was not displaying any change in composer.json.

Issue #1316, opened 2022-08-27
Target directory is not created


Flower7C3:
In 
https://github.com/dustin10/VichUploaderBundle/blob/6b428642694cbfeb1be6e956ba59382d89ae1ee6/src/Storage/FileSystemStorage.php#L25 the target directory is not created when I use `ReplacingFile` (like described in https://github.com/dustin10/VichUploaderBundle/blob/master/docs/other_usages/replacing_file.md#inject-files-from-other-sources)

gdbonino:
The same for me

garak:
@UlrichThomasGabor do you have an idea about this?

UlrichThomasGabor:
According to a php comment `move_uploaded_file` won't create a directory also, therefore I guess this is not related to the new `ReplacingFile` functionality, but should be a general problem.

What is your `vich_uploader.yaml` content and what is the existing directory structure? I guess you tried to use one of the [directory namers](https://github.com/dustin10/VichUploaderBundle/blob/master/docs/namers.md#directory-namer)?

UlrichThomasGabor:
Should be fixed via #1318

PR #1318, opened 2022-09-05
Fixes issue #1316

UlrichThomasGabor:
I guess the `FileSystemStorage` just lacked functionality for `DirectoryNamerInterface`/subdirectories. Don't know how this issue has not come up earlier...

### DIFF
--- a/tests/Storage/FileSystemStorageTest.php
+++ b/tests/Storage/FileSystemStorageTest.php
@@ -264,11 +264,13 @@ final class FileSystemStorageTest extends StorageTestCase
      */
     public function testUploadedFileIsCorrectlyMoved(string $uploadDir, string $dir, string $expectedDir): void
     {
+        $uploadDir = $this->root->url().\DIRECTORY_SEPARATOR.$uploadDir;
+        $expectedDir = $this->root->url().\DIRECTORY_SEPARATOR.$expectedDir;
         $file = $this->getUploadedFileMock();
 
         $file
             ->method('getClientOriginalName')
-            ->willReturn('filename.txt');
+            ->willReturn('test.txt');
 
         $this->mapping
             ->expects(self::once())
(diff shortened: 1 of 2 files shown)


tests/Storage/FileSystemStorageTest.php
```diff
@@ -341,24 +342,63 @@ public function testReplacingFileIsCorrectlyUploaded(): void
...
+        $this->storage->upload($this->object, $this->mapping);
+    }
+
     public function filenameWithDirectoriesDataProvider(): array
     {
         return [
             // upload dir, dir, expected dir
             [
-                '/root_dir',
+                '/storage/vich_uploader_bundle',
```
garak, line 389: can we provide useful keys, here instead of relying on numeric keys? E.g. something like "directory with subdirectory" and so on

garak (changes requested):

tests/Storage/FileSystemStorageTest.php
UlrichThomasGabor: No, this is a [`DataProvider`](https://phpunit.de/manual/3.7/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers) for the [unit test above](https://github.com/dustin10/VichUploaderBundle/blob/1d891f943b2cd6fe15b5befe572cd5b78a2fcf4f/tests/Storage/FileSystemStorageTest.php#L262), and I do not see anything in the documentation that this functionality supports named parameters.

garak: I mean names for data, not for parameters. See https://phpunit.readthedocs.io/en/9.5/writing-tests-for-phpunit.html?#writing-tests-for-phpunit-data-providers-examples-datatest1-php

UlrichThomasGabor: changed

garak: Thanks. What about omitting "upload with "? It looks like labels like "two subdirectories" are enough to understand the case

UlrichThomasGabor: Yes, changed again

garak (approved):

PR #1320, opened 2022-09-14
docs: add some warnings regarding the nullable type variable for enti…

TheoSrz69:
…ty properties

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -71,6 +71,10 @@ The `UploadableField` annotation has a few options. They are as follows:
 * `originalName`: the property that will contain the original name of the uploaded file.
 * `dimensions`: the property that will contain the dimensions of uploaded **image file**
 
+_Warning__
+Be sure that the return types (for getter methods) & the parameter types (for setter methods) use the "?" symbol.
+It is mandatory for removing correctly Symfony entity, with no error.
+
 Let's look at an example using a fictional `Product` ORM entity:
 
 ``` php


docs/usage.md
```diff
@@ -71,6 +71,26 @@ The `UploadableField` annotation has a few options. They are as follows:
 * `originalName`: the property that will contain the original name of the uploaded file.
 * `dimensions`: the property that will contain the dimensions of uploaded **image file**
 
+_Warning__ : all these entity properties, in their respective getter & setter methods,  should accept null value.
+It is mandatory for removing correctly entity with symfony. For that, simply add an "?" symbol before the return/send variable type.
```
garak, line 75: The documentation is already showing the proper "?" symbols, so I guess we should suggest, instead, to ensure that such symbols are present (if the user followed the examples, they should have already added them).
Moreover, "Symfony" should be capitalized.

garak (changes requested):

docs/usage.md
TheoSrz69: You're (partially) right. The example is good, but it's only an example. Lot of people would not pay attention to detail if document doesn't mention it. I reduced the explanation for being more brief.

garak: Still a couple of issues: I see a double underscore after "Warning", I guess it's a typo; and we should highlight the newly added paragraph (see create_a_custom_directory_namer.md for an example)

garak: bump

garak:
Closing.
Feel free to open a new PR if relevant

PR #1322, opened 2022-10-14
urlencode filename in resolve uri

pbreteche:
Filenames with non URL friendly chars break links

For example:
filename = 'training-C#-overview.pdf'
will create link as:
https://example.com/path/prefix/training-C#-overview.pdf
instead of:
https://example.com/path/prefix/training-C%23-overview.pdf

URL encoding should prevent browsers from interpreting this char.


garak:
This is not a good solution, since the sanitation is done during visualization (multiple times) instead of during upload (only once).
The bundle is already providing options to achieve that, via namers

PR #1324, opened 2022-10-30
[Commands] Fix Symfony 6.1 deprecations

asuri0n:
The `$defaultName` property in commands has been [deprecated](https://github.com/symfony/symfony/commit/ca3458c8781eafd9ff40bac50a254334e526dc10) in Symfony 6.1

Commands in Symfony 6 should use the `AsCommand` attribute instead.

Tests command OK.

### DIFF
--- a/src/Command/MappingDebugClassCommand.php
+++ b/src/Command/MappingDebugClassCommand.php
@@ -17,11 +17,6 @@ final class MappingDebugClassCommand extends Command
         parent::__construct();
     }
 
-    public static function getDefaultName(): string
-    {
-        return 'vich:mapping:debug-class';
-    }
-
     protected function configure(): void
     {
         $this
--- a/src/Command/MappingDebugClassCommand.php
+++ b/src/Command/MappingDebugClassCommand.php
@@ -2,6 +2,7 @@
 
 namespace Vich\UploaderBundle\Command;
 
+use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Completion\CompletionInput;
 use Symfony\Component\Console\Completion\CompletionSuggestions;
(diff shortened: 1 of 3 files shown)


garak:
Where is `$defaultName`?

garak:
> I meant method `getDefaultName()` and property `$defaultName` are deprecated.

But, again, we don't use `$defaultName`.
About the deprecation of `getDefaultName()`, can you provide a reference?

asuri0n:
I've linked the commit for the reference. Precisely [here](https://github.com/symfony/symfony/commit/ca3458c8781eafd9ff40bac50a254334e526dc10#diff-af3c198cbbaa2b1b7075232bd660a286f94dafa023c362d0e67315369554c38fR83).

And this deprecation info in logs when using your bundle :
> deprecation.INFO: User Deprecated: Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute to the "Vich\UploaderBundle\Command\MappingDebugCommand" class instead. {"exception":"[object] (ErrorException(code: 0): User Deprecated: Since symfony/console 6.1: Relying on the static property \"$defaultName\" for setting a command name is deprecated. Add the \"Symfony\\Component\\Console\\Attribute\\AsCommand\" attribute to the \"Vich\\UploaderBundle\\Command\\MappingDebugCommand\" class instead.

Is it what you are waiting for as a "reference" ?

garak:
Unfortunately, that's not a reference. The method is not mentioned.
You're probably using an old version of the bundle.

garak:
Thanks for proposing a PR.
Next time, it would be better to open an issue first.

asuri0n:
We are using  1.21.1 version.
I didn't imagine this, but I'll double-check. sorry for bothering you

PR #1331, opened 2022-11-14
Hint for annotation usage in UPGRADE.md

virtualize:
It took me quite some time to figure auf why our mappings are not working anymore. So I've added some help for users who want to upgrade but are still using annotations. I think this will save others some time.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -6,6 +6,6 @@
 * all methods have now return types
 * all constructors now use property promotion
 * all deprecated features were removed
-* [BC BREAK] the new default type for mapping is "attribute". You can still use annotations, but you need an explicit definition (set "annotation" as value for vich_uploader.metadata.type config key)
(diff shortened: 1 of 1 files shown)


UPGRADE.md
```diff
@@ -6,6 +6,6 @@
 * all methods have now return types
 * all constructors now use property promotion
 * all deprecated features were removed
-* the new default type for mapping is "attribute". You can still use annotations, but you need an explicit definition.
+* [BC BREAK] the new default type for mapping is "attribute". You can still use annotations, but you need an explicit definition (set "annotation" as value for vich_uploader.metadata.type config key)
```
garak, line 9: every single point in this list is a BC break, so please remove the tag.
Moreover, the key config name must be quoted.

garak (changes requested):

UPGRADE.md
virtualize: Done. Thanks for your fast response.

garak (approved):

PR #1333, opened 2022-11-15
Remove lazy flag from GaufretteStorage

bobvandevijver:
Classes that are final cannot be marked as lazy.

This service was originally marked as lazy with https://github.com/dustin10/VichUploaderBundle/pull/630, but I cannot reproduce a circular reference with my project after removing the lazy flag.


garak (approved):

garak:
Thank you @bobvandevijver

PR #1334, opened 2022-11-22
XSD schema for XML mapping

yivi:
Provides an XSD schema so that creating XML mapping is easier, since IDE can use it for autocompletion

### Improvement

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

Creating an XML document without an associated XSD schema is error prone. With the provided XSD schema, IDEs can provide autocompletion, and XML files can be validated.

### DIFF
--- a/docs/mapping/xml.md
+++ b/docs/mapping/xml.md
@@ -6,7 +6,8 @@ format and comes with the following syntax to declare your uploadable fields:
 ```xml
 <!-- config/vich_uploader/Entity.Product.xml -->
 <!-- Attributes "mapping", "name" and "filename_property" are required -->
-<vich_uploader class="Acme\DemoBundle\Entity\Product">
+<vich_uploader xmlns="https://vich-uploader-bundle/schema/"
+               class="Acme\DemoBundle\Entity\Product">
     <field mapping="products" name="imageFile" filename_property="imageName"
            size="imageSize" dimensions="imageDimensions" mime_type="imageMimeType" original_name="imageOriginalName" />
 </vich_uploader>
--- a/vich_uploader.xsd
+++ b/vich_uploader.xsd
@@ -1,0 +1,32 @@
+<xs:schema
+        attributeFormDefault="unqualified"
+        elementFormDefault="qualified" xmlns:xs="http://www.w3.org/2001/XMLSchema"
+        targetNamespace="https://vich-uploader-bundle/schema/"
+>
+    <xs:element name="vich_uploader">
+        <xs:complexType>
+
+            <xs:sequence>
+
+                <xs:element name="field">
+                    <xs:complexType>
+                        <xs:simpleContent>
+                            <xs:extension base="xs:string">
+                                <xs:attribute type="xs:string" name="name" use="required"/>
+                                <xs:attribute type="xs:string" name="mapping" use="required"/>
+                                <xs:attribute type="xs:string" name="filename_property" use="required"/>
+                                <xs:attribute type="xs:string" name="size"/>
+                                <xs:attribute type="xs:string" name="dimensions"/>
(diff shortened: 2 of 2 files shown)


vich_uploader.xsd
```diff
@@ -0,0 +1,32 @@
...
+
+            <xs:sequence>
+
+                <xs:element name="field">
+                    <xs:complexType>
+                        <xs:simpleContent>
+                            <xs:extension base="xs:string">
+                                <xs:attribute type="xs:string" name="name" use="required"/>
+                                <xs:attribute type="xs:string" name="mapping" use="required"/>
+                                <xs:attribute type="xs:string" name="filename_property" use="required"/>
```
garak, line 17: I don't use XML mapping very much, but if I remember correctly the `filename_property` should be optional, it should work without it when you have a simple mapping with only a single property.
Could you check it and, in case, remove the `required` property?

garak (changes requested):

vich_uploader.xsd
yivi: I think that thhe `filename_property` is required. On the docs it's stated on all mappings as required  (and makes sense it's required, IMO).

It's mentioned [here](https://github.com/dustin10/VichUploaderBundle/blob/master/docs/usage.md#step-2-link-the-upload-mapping-to-an-entity) for attribute mapping, [here](https://github.com/dustin10/VichUploaderBundle/blob/master/docs/mapping/xml.md) for XML mapping, and [here](https://github.com/dustin10/VichUploaderBundle/blob/master/docs/mapping/yaml.md) for YAML mapping.

garak: It could be a case of outdated docs. At one point, we introduced that possibility and maybe we forgot to check for all documentation.
Anyway, can you try and tell me if it works without it?

yivi: Finally got around to try it
.
IMO, it's a rather weird situation.

The `filename_property` config is _not_ mandatory, **but** it will only work if the entity has a an accessible property called exactly `file_namep`.

E.g. if you do not use that configuration, but your entity does not have this property or uses any slightly different name, the upload will fail:

imagen


Since having the property **is** required, I think it makes sense to leave the requirement to configure the mapping correctly, as stated on the existing docs.

Less errors this way.

yivi: Making it non-required would make necessary to make many more changes on the docs, and explain that this will work if you have an specific name for the this property, which would open the door for more user errors.

garak (approved):

garak:
Just another little doubt: should we update the [docs](https://github.com/dustin10/VichUploaderBundle/blob/master/docs/mapping/xml.md) as well?

yivi:
sorry, I was out on holidays.

I'll update the XML in the docs as well, so users know they can benefit from the schema.

docs/mapping/xml.md
```diff
@@ -6,7 +6,8 @@ format and comes with the following syntax to declare your uploadable fields:
 ```xml
 <!-- config/vich_uploader/Entity.Product.xml -->
 <!-- Attributes "mapping", "name" and "filename_property" are required -->
-<vich_uploader class="Acme\DemoBundle\Entity\Product">
+<vich_uploader xmlns="https://vich-uploader-bundle/schema/"
```
garak, line 9: So, I know that the "xmlns" attribute is not required, but I'm wondering how can the IDE understand where it's located (and so providing autocompletion). Could you clarify that?

yivi: At least on phpstorm, if the file is within the project directories it will be picked up automatically.

garak:
Thank you @yivi

PR #1336, opened 2022-12-01
Remove ext-mongo polyfill

alcaeus:
With Doctrine MongoDB ODM 1.x no longer being supported, this library no longer relies on the deprecated `ext-mongo` extension, so we no longer need to install the polyfill for it.


garak (approved):

garak:
thank you @alcaeus

PR #1338, opened 2022-12-12
deprecate: spitNameByExtension, use splitNameByExtension instead

ockstadt:


src/Util/FilenameUtils.php
```diff
@@ -19,12 +23,26 @@ private function __construct()
...
+     *
+     * @return array An array of basename and extension
+     *
+     * @deprecated, use splitNameByExtension() instead
+     */
+    public static function spitNameByExtension(string $filename): array
+    {
+        trigger_deprecation('vich/uploader-bundle', '2.0.2', '"%s()" is deprecated, use "splitNameByExtension()" instead.', __METHOD__);
+
+        return $this->splitNameByExtension($filename);
```
garak, line 46: Method is static

src/Util/FilenameUtils.php
```diff
@@ -19,12 +23,26 @@ private function __construct()
...
+    /**
+     * Splits filename for array of basename and extension.
+     *
+     * @return array An array of basename and extension
+     *
+     * @deprecated, use splitNameByExtension() instead
+     */
+    public static function spitNameByExtension(string $filename): array
+    {
+        trigger_deprecation('vich/uploader-bundle', '2.0.2', '"%s()" is deprecated, use "splitNameByExtension()" instead.', __METHOD__);
```
garak, line 44: We don't add deprecation in patch versions. Please change this with `2.1.0`

garak (changes requested):

src/Util/FilenameUtils.php
ockstadt: Thank you, fixed it.

src/Util/FilenameUtils.php
ockstadt: Changed to 2.1.0 as requested.

garak: On second thought, I realized that we don't need to deprecate this method, since the class is marked as "internal".
By the way, the class itself looks a bit overkill, since it's used only in our `Transliterator` (that's internal too).
So, please change this PR (or open a new one, if you prefer) to remove the class and move the method as a private method of the `Transliterator` class

ockstadt: Don't you want to give users the chance to remove this function call in case they just ignored that it is marked as "internal"? I could then change the comment to state that this function will be removed in the next major release?

garak: I don't want to wait for the next major release. We'll add a note about the breaking change in the UPGRADE file: if someone is using the internal class, it's up to them to use caution with upgrades.

ockstadt: I moved splitNameByExtension to /src/Util/Transliterator.php updated the UPGRADE.md, removed /src/Util/FilenameUtils.php and /tests/Util/FilenameUtilsTest.php.

src/Util/Transliterator.php
```diff
@@ -18,12 +21,26 @@ public function __construct(private readonly SluggerInterface $slugger)
      */
     public function transliterate(string $string, string $separator = '-'): string
     {
-        [$filename, $extension] = FilenameUtils::spitNameByExtension($string);
+        [$filename, $extension] = $this->splitNameByExtension($string);
         $transliterated = $this->slugger->slug($filename, $separator);
         if ('' !== $extension) {
-            $transliterated .= '.'.$extension;
+            $transliterated .= '.' . $extension;
```
garak, line 27: This change breaks our coding standard

UPGRADE.md
```diff
@@ -1,3 +1,7 @@
+# Upgrading from v2.0.1 to v2.1.0
+
+* the internal class src/Util/FilenameUtils.php has been removed.
```
garak, line 3: the name of the class should be like `this`

garak (changes requested):

src/Util/Transliterator.php
ockstadt: Adapted file to match coding standard.

UPGRADE.md
ockstadt: Adapted documentation accordingly.

garak (approved):

garak:
thank you @ockstadt

PR #1339, opened 2022-12-12
Fix column span in mapping collector table header

endelwar:


garak (approved):

garak:
Good catch! Thank you @endelwar

PR #1346, opened 2023-01-05
Check metadata type when adding annotation/attribute mapping drivers

### LINKED ISSUE #1344: Broken metadata when updated Doctrine to 2.8.0
### Bug Report

|    Q            |   A
|---------------- | ------
| BC Break        | no
| Bundle version  |  2.0.1
| Symfony version | 6.2.2
| PHP version     | 8.2.0

#### Summary

`Vich\UploaderBundle\Exception\NotUploadableException` on all entities after update Doctrine to 2.8.0

#### Current behavior

All the uploadable entities are broken with `Vich\UploaderBundle\Exception\NotUploadableException`

doctrine/doctrine-bundle doesn't require `doctrine/annotations` anymore which breaks metadata reader, to fix it I've added `doctrine/annotations:1.14.1` to dependencies list, but it seems not to be a correct solution. It'd be nice if we can drop requirement of deprecated library.

#### How to reproduce

composer require doctrine/doctrine-bundle:2.8.0 and refer to any uploadable field of any uploadable entity.

#### Expected behavior

uploadable entities works.

andyexeter:
Fixes #1344

The `annotation_reader` service is _not_ required for attributes to be read, but if the `doctrine/annotations` package isn't present the service does not exist meaning attribute metadata is not read.

This PR resolves this issue by adding the meta data type config as a parameter and modifying the check for the existence of the service so it short-circuits if the meta data type is `attribute`.

### DIFF
--- a/src/DependencyInjection/Compiler/RegisterMappingDriversPass.php
+++ b/src/DependencyInjection/Compiler/RegisterMappingDriversPass.php
@@ -18,7 +18,9 @@ final class RegisterMappingDriversPass implements CompilerPassInterface
             new Reference('vich_uploader.metadata_driver.xml'),
         ];
 
-        if ($container->has('annotation_reader')) {
+        $metaDataType = $container->getParameter('vich_uploader.metadata_type');
+
+        if ('attribute' === $metaDataType || ('annotation' === $metaDataType && $container->has('annotation_reader'))) {
             $managers = [];
             if ($container->hasDefinition('doctrine_mongodb')) {
                 $managers[] = new Reference('doctrine_mongodb');
--- a/src/DependencyInjection/VichUploaderExtension.php
+++ b/src/DependencyInjection/VichUploaderExtension.php
@@ -38,6 +38,7 @@ final class VichUploaderExtension extends Extension
         // define a few parameters
         $container->setParameter('vich_uploader.default_filename_attribute_suffix', $config['default_filename_attribute_suffix']);
         $container->setParameter('vich_uploader.mappings', $config['mappings']);
+        $container->setParameter('vich_uploader.metadata_type', $config['metadata']['type']);
 
         if (\str_starts_with((string) $config['storage'], '@')) {
             $container->setAlias('vich_uploader.storage', \substr((string) $config['storage'], 1));


src/DependencyInjection/Compiler/RegisterMappingDriversPass.php
```diff
@@ -18,7 +18,9 @@ public function process(ContainerBuilder $container): void
             new Reference('vich_uploader.metadata_driver.xml'),
         ];
 
-        if ($container->has('annotation_reader')) {
+        $metaDataType = $container->getParameter('vich_uploader.metadata_type');
+
+        if ('attribute' === $metaDataType || ('annotation' === $metaDataType && $container->has('annotation_reader'))) {
```
garak, line 23: I don't see a case in which this condition could be false: the metadata type defaults to "attribute" and it can be only "attribute" or "annotation", the container has always "annotation_reader"

garak (changes requested):

src/DependencyInjection/Compiler/RegisterMappingDriversPass.php
andyexeter: @garak I thought it could be `xml` or `yaml` as well?

garak: OK, sorry, you're right, the presence of "annotation_reader" is not ensured.
But so at this point, I don't get why you need to load the managers into the `vich_uploader.metadata_driver.annotation` service. If they're only needed when using annotations and you're using attributes, you should be OK without them

andyexeter: I'm really not sure - I was trying to make a minimal change to fix the issue at hand and keep the status quo, rather than making more wholesale changes.

wimwinterberg:
When doctrine/annotations is not installed the interface (Doctrine\Common\Annotations\Reader) used by "Metadata/Driver/AttributeReader.php" also does not exists. So code will fail here as well.

I am also looking for a way for it to work without doctrine/annotations

andyexeter:
@wimwinterberg you are right - I didn't realise `doctrine/annotations` was still required by a dev dependency so didn't spot that during my testing.

I'm going to close this PR as the approach to solve the issue is incorrect.

PR #1351, opened 2023-01-30
:pencil: (Doctrine) Update docs & package suggestions

feymo:
As discussed in #1344 , before finding a proper way to deal with this issue, i propose to update the documentation & package suggestion.
This way, developpers encoutering this issue will have an easy way to solve it until we implement a more independant solution

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,6 @@
         "ext-sqlite3": "*",
         "dg/bypass-finals": "^1.3",
         "doctrine/doctrine-bundle": "^2.7",
-        "doctrine/mongodb-odm": "^2.4",
         "doctrine/orm": "^2.13",
         "knplabs/knp-gaufrette-bundle": "dev-master",
         "league/flysystem-bundle": "^2.3",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -33,7 +33,6 @@
         "ext-sqlite3": "*",
         "dg/bypass-finals": "^1.3",
         "doctrine/doctrine-bundle": "^2.7",
-        "doctrine/mongodb-odm": "^2.4",
```
garak, line 36: this change is unrelated and should be reverted

garak (changes requested):

composer.json
feymo: Totally, i removed this package by mistake and forgot to put it back... Thanks ! 👍🏼

garak (changes requested):
Please check failures related to the documentation format

garak (changes requested):
I'm not sure if you actually checked the lint errors...
In any case, here they are:
```
docs/known_issues.md:138:103 MD009/no-trailing-spaces Trailing spaces [Expected: 0 or 2; Actual: 1]
docs/known_issues.md:140:98 MD009/no-trailing-spaces Trailing spaces [Expected: 0 or 2; Actual: 1]
```

garak (approved):

garak:
Thank you @feymo !

PR #1357, opened 2023-02-05
do not rely on annotations to use attributes

### LINKED ISSUE #1344: Broken metadata when updated Doctrine to 2.8.0

garak:
Fix #1344

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,7 @@
         "symfony/yaml": "^5.4 || ^6.0"
     },
     "suggest": {
-        "doctrine/annotations": "If your project use doctrine-bundle:>=2.8, this package is required to use annotations or attributes",
+        "doctrine/annotations": "If you use doctrine/doctrine-bundle >2.7, this package is required to use annotations",
         "doctrine/doctrine-bundle": "For integration with Doctrine",
         "doctrine/mongodb-odm-bundle": "For integration with Doctrine ODM",
         "doctrine/orm": "For integration with Doctrine ORM",
--- a/src/DependencyInjection/VichUploaderExtension.php
+++ b/src/DependencyInjection/VichUploaderExtension.php
@@ -11,6 +11,7 @@ use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Extension\Extension;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
+use Vich\UploaderBundle\Exception\MissingPackageException;
 use Vich\UploaderBundle\Metadata\CacheWarmer;
 use Vich\UploaderBundle\Storage\StorageInterface;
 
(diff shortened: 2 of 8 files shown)


src/DependencyInjection/VichUploaderExtension.php
```diff
@@ -4,13 +4,15 @@
 
 use Doctrine\Common\Annotations\AnnotationReader;
 use Symfony\Component\Config\FileLocator;
+use Symfony\Component\CssSelector\Parser\Reader;
```
andyexeter, line 7: Wrong class imported here. Should be `Doctrine\Common\Annotations\Reader`.

composer.json
```diff
@@ -58,7 +59,7 @@
         "symfony/yaml": "^5.4 || ^6.0"
     },
     "suggest": {
-        "doctrine/annotations": "If your project use doctrine-bundle:>=2.8, this package is required to use annotations or attributes",
+        "doctrine/annotations": "If you use doctrine/doctrine-bundle >2.7, this package is required to use annotations",
```
andyexeter, line 62: Incorrect wording here - the package is required to use annotations _and_ attributes - not just annotations.

src/DependencyInjection/VichUploaderExtension.php
```diff
@@ -138,16 +140,22 @@ protected function registerAnnotationStrategy(ContainerBuilder $container, array
         }
 
         switch ($config['metadata']['type']) {
-            case 'attribute':
+            case 'annotation':
```
andyexeter, line 143: This does not resolve the issue because this bundle's attribute reader requires `Doctrine\Common\Annotations\Reader` which is part of `doctrine/annotations`.

With this PR, when using attributes the following error is now thrown in src/Metadata/Driver/AttributeReader.php (line 15):

Attempted to load interface "Reader" from namespace "Doctrine\Common\Annotations".
Did you forget a "use" statement for "Symfony\Component\CssSelector\Parser\Reader"?


Screenshot from 2023-02-06 12-35-20

composer.json
garak: the purpose of this PR is indeed allowing the use of attributes without relying on doctrine annotations

src/DependencyInjection/VichUploaderExtension.php
garak: My bad, fixed

src/DependencyInjection/VichUploaderExtension.php
garak: It should be working now

garak: > This seems to be working for attributes +1

thank you for checking. Do you have the possibility to check it for annotations too? It would be awesome

PR #1365, opened 2023-02-18
Allow use of array to resolve uri

### LINKED ISSUE #1361: Use an array instead of an object
|    Q            |   A
|---------------- | ------
| Bundle version  | 2.0.1
| Symfony version | 6.2.6
| PHP version     | 8.2.2

### Support Question

Hello. I am currently running into an issue that could be solved easily by just reading the documentation, but the documentation is outdated.

I want to generate a path (preferably in a twig template) to my image using an array instead of an object. This is because I am using Meilisearch and for performance reasons, I only get results as an array.

In previous versions, it seemed possible to pass in an array and the class name as a third argument, as [stated in the doc](https://github.com/dustin10/VichUploaderBundle/blob/master/docs/generating_urls.md#generating-a-url-in-a-twig-template). However, it was marked as deprecated in earlier versions and later removed. Is there still any support for this? Or do I absolutely need to pass in an entity?

YummYume:
Closes #1361.

This PR (re)allows the use of an array instead of an object to resolve a file path. I also went ahead and reintroduced this feature for the twig filter.

I moved the `fixPropertyPath` function to a utils file since it was also needed in the `directoryName` function because of the date property contained in the array.

### DIFF
--- a/src/Mapping/PropertyMapping.php
+++ b/src/Mapping/PropertyMapping.php
@@ -127,8 +127,8 @@ final class PropertyMapping
      *
      * @internal
      *
-     * @param object $obj      The object or array from which read
-     * @param string $property The property to read
+     * @param object|array $obj      The object or array from which read
+     * @param string       $property The property to read
      *
      * @throws \InvalidArgumentException
      */
--- a/src/Mapping/PropertyMapping.php
+++ b/src/Mapping/PropertyMapping.php
@@ -7,6 +7,7 @@ use Symfony\Component\PropertyAccess\PropertyAccess;
 use Symfony\Component\PropertyAccess\PropertyAccessor;
 use Vich\UploaderBundle\Naming\DirectoryNamerInterface;
 use Vich\UploaderBundle\Naming\NamerInterface;
+use Vich\UploaderBundle\Util\PropertyPathUtils;
 
 /**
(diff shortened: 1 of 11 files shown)


src/Util/PropertyPathUtils.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Vich\UploaderBundle\Util;
+
+abstract class PropertyPathUtils
```
garak, line 5: This class should be final instead, we don't want to allow extensions

src/Mapping/PropertyMapping.php
```diff
@@ -82,13 +83,13 @@ public function setFile(object $obj, File $file): void
     /**
      * Gets the fileName property of the given object.
      *
-     * @param object $obj The object
+     * @param object $obj The object or array
```
garak, line 86: The phpdoc type should be "object|array" (here and in all other points where is relevant)

garak (changes requested):

YummYume:
Thanks @garak, fixed

garak:
> Thanks @garak, fixed

Actually not, please check again

YummYume:
My bad, should be good now :ok_hand:

garak (approved):

PR #1371, opened 2023-03-07
[v2] Fix MongoDB changeset

IonBazan:
Ports over changes from #1277 to `master` branch (v2).


src/Adapter/ORM/DoctrineORMAdapter.php
```diff
@@ -14,19 +16,11 @@
 final class DoctrineORMAdapter implements AdapterInterface
 {
     /**
-     * @param \Doctrine\ORM\Event\LifecycleEventArgs $event
+     * @param LifecycleEventArgs $event
```
IonBazan, line 19: I wonder if this could be replaced with `assert($event instanceof LifecycleEventArgs)` in the method body.

PR #1373, opened 2023-03-10
StorageInterface resolveStream nullable $fieldName

pauliuspetronis:


garak (changes requested):
An explanation would be much appreciated

pauliuspetronis:
Other 2 functions (resolvePath and resolveUri) accept fieldName as nullable string. Why resolveStream function makes different? I think it like a mistake or misstype. If you have one "uploadable" field per entity - you have to pass field name every time only on this function.

src/Storage/StorageInterface.php
```diff
@@ -56,10 +56,10 @@ public function resolveUri(object|array $obj, ?string $fieldName = null, ?string
...
      *
      * @param object|array $obj       The object
-     * @param string       $fieldName The field to use
+     * @param string|null  $fieldName The field to use
      * @param string|null  $className The object's class. Mandatory if $obj can't be used to determine it
      *
      * @return resource|null The resolved resource or null if file not stored
      */
-    public function resolveStream(object|array $obj, string $fieldName, ?string $className = null);
+    public function resolveStream(object|array $obj, ?string $fieldName = null, ?string $className = null);
```
garak, line 64: This is a BC break, as any changes in an interface.
But I understand that it's currently inconsistent, and it's not a big change.
Anyway, we need to annotate this change in the UPGRADE file

garak (changes requested):

src/Storage/StorageInterface.php
pauliuspetronis: Done

UPGRADE.md
```diff
@@ -1,3 +1,7 @@
+# Upgrading from v2.1.1 to v2.2.0
```
garak, line 1: This doesn't seem right. I understand that you were misled by the other entry (which is also wrong), but specifying the patch version here doesn't make sense.
Please change it to "from 2.1 to 2.2". Please change the other one accordingly ("from 2.0 to 2.1").

garak (changes requested):

UPGRADE.md
```diff
@@ -1,8 +1,12 @@
-# Upgrading from v2.0.1 to v2.1.0
+# Upgrading from v2.1 to v2.2
+
+* The signature of `StorageInterface::resolveStream` method was changed. Nullable value of parameters `$fieldName`.
```
garak, line 3: `$fieldName` is a single parameter, right?
Anyway, this sentence is not clear, please reword it as follows: "The `$fieldName` parameter is now nullable."

garak (changes requested):

UPGRADE.md
pauliuspetronis: Fixed. Sorry for my en.

garak (approved):

PR #1379, opened 2023-03-31
Explicit that nullable is a necessity

mysterty:
I just experienced this issue. I didn't thinck that "nullable" was REALLY needed, as i was using my own uploading before and set my field as "uniq". Maybe i wont be the only person experiencing it so… i guess it cost nothing to precise it.

more here : https://stackoverflow.com/questions/75861160/symfony-6-vichuploader-cant-update-delete-file/75902101#75902101

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -97,6 +97,7 @@ class Product
     #[Vich\UploadableField(mapping: 'products', fileNameProperty: 'imageName', size: 'imageSize')]
     private ?File $imageFile = null;
 
+    // NOTE: be carefull, this field needs to be nullable.
     #[ORM\Column(nullable: true)]
     private ?string $imageName = null;
 


garak:
Thank you for your contribution.
In any Doctrine mapping, every property that has a nullable type must be declared as nullable. This is not related to this bundle, but to the way Doctrine manages the mapping itself.

So, I think it should be understood by anyone, and anyway the documentation is already proposing the correct way.

mysterty:
It's not obvious that the property needs to be declared nullable.

As i read the doc, it was just for the exemple that the property is nullable.

For database integrity, we could conciderate that the upload field needs to be not null. Indeed, an uploadable entity without data attached should be destroyed…

garak:
If you declare a property with a default value of `null` (like in the example you linked), you must declare it as `nullable`.
Maybe it's not obvious (one could argue that Doctrine should be able to understand it), but it's how Doctrine works.

garak:
It's not needed to be explicit: the `null` default value is already provided by the example (and even by generated code, in the case of the maker).
The `nullable` mapping is a direct consequence of the property being `null`, and not setting it violates the Doctrine standard, not a rule set in this bundle.

mysterty:
In absolute, i would agree with you. But symfony allow to map a doctrine not nullable property with a nullable php property **AND do it _by default_**

So, concidering this fact, the actual documentation is ambigous.

garak:
> In absolute, i would agree with you. But symfony allow to map a doctrine not nullable property with a nullable php property **AND do it _by default_**

If so, this is a bug in Symfony itself (I guess in the maker bundle)

PR #1381, opened 2023-04-05
Enable error bunbbling in the embebbed file field

Safranil:
Without the `error_bubbling` in the embedded file field, any error from the field doesn't get displayed in the form.

In my case, I get a size limit that is hidden by VichUploader in the form but the field is marked as invalid by Symfony Form component.

I'm using the Symfony backed Bootstrap theme and i also got the same behavior without any theme.

Note: The error appear when the file size is greater than `upload_max_filesize`.

image
image

About my environment:
- Symfony 5.4.18
- PHP 8.1.17
- vich/uploader-bundle 2.1.1

### DIFF
--- a/src/Form/Type/VichFileType.php
+++ b/src/Form/Type/VichFileType.php
@@ -61,6 +61,7 @@ class VichFileType extends AbstractType
             'label' => $options['label'],
             'attr' => $options['attr'],
             'translation_domain' => $options['translation_domain'],
+            'error_bubbling' => true,
         ]);
 
         $builder->addModelTransformer(new FileTransformer());


garak:
I'm sorry but I can't reproduce your problem.
If you still think it's something wrong with this bundle, please open an issue providing a way to reproduce it.
Thank you.

Safranil:
I still have the issue with a simple case (only one field with Vich), you can see the code here:
https://cloud.safranil.fr/s/eLkNj8YXPTS7T4f

Once you set `upload_max_filesize` to `2M` and upload the file `33771058105_af4685ee0e_5k.jpg` (file present in archive for the reproduction), the error is then added to the file field but not displayed in the page until the bubbling is enabled.

Issue #1388, opened 2023-05-31
Doctrine subscribers as services is deprecated, declare listeners instead


laurentmuller:
### BC Break Report

I have update Symfony to version 6.3.0. A deprecation is now raised.

|    Q        |   A
|------------ | ------
| BC Break    | No
| Version     | 2.1.1

#### Summary

When using this bundle, the following deprecation is raised:

`Since symfony/doctrine-bridge 6.3: Using Doctrine subscribers as services is deprecated, declare listeners instead`

The deprecation is raised by the `Vich\UploaderBundle\EventListener\Doctrine\BaseListener` class. This class implement the `EventSubscriber` interface.

Please see [#49918](https://github.com/symfony/symfony/pull/49918) and replace with event listener.

#### Previous behavior

No deprecation.

PR #1391, opened 2023-06-16
Replaced EventSubscriber by listener.

laurentmuller:
This is the change for issue #1388.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,6 @@
     "require": {
         "php": "^8.1",
         "ext-simplexml": "*",
-        "doctrine/annotations": "^2.0",
         "doctrine/persistence": "^3",
         "jms/metadata": "^2.4",
         "symfony/config": "^5.4 || ^6.0",
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,7 @@
     "require": {
         "php": "^8.1",
         "ext-simplexml": "*",
+        "doctrine/annotations": "^2.0",
         "doctrine/persistence": "^3",
         "jms/metadata": "^2.4",
         "symfony/config": "^5.4 || ^6.0",
--- a/src/Adapter/ORM/DoctrineORMAdapter.php
+++ b/src/Adapter/ORM/DoctrineORMAdapter.php
@@ -22,7 +22,7 @@ final class DoctrineORMAdapter implements AdapterInterface
     {
         $object = $event->getObject();
 
-        $em = $event->getObjectManager();
+        $em = $event->getEntityManager();
         $uow = $em->getUnitOfWork();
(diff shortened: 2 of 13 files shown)


composer.json
```diff
@@ -17,6 +17,7 @@
     "require": {
         "php": "^8.1",
         "ext-simplexml": "*",
+        "doctrine/annotations": "^2.0",
```
garak, line 20: I can't see any use of annotations... Do we really need to re-introduce this?

garak (changes requested):

composer.json
laurentmuller: This is used by the `VichUploaderExtension` class on function `registerAnnotationStrategy`. If not, a `MissingPackageException` is raised:

`Vich\UploaderBundle\Exception\MissingPackageException : Annotations support missing. Try running "composer require doctrine/annotations".`

garak: That's a conditional requirement, which is already covered in our "suggest" section. Please remove it from "require"

laurentmuller: OK.

garak:
I guess the test failure depends on some lower dependencies that need to be raised

src/Adapter/ORM/DoctrineORMAdapter.php
```diff
@@ -22,7 +22,7 @@ public function recomputeChangeSet(BaseLifecycleEventArgs $event): void
     {
         $object = $event->getObject();
 
-        $em = $event->getEntityManager();
+        $em = $event->getObjectManager();
```
garak, line 25: This change is about a different issue, it should be in another PR

laurentmuller: You are right. How we can move it to the #1393 PR?

garak (approved):

src/DependencyInjection/VichUploaderExtension.php
```diff
@@ -196,7 +196,7 @@ protected function registerListeners(ContainerBuilder $container, array $config)
     {
         $servicesMap = [
             'inject_on_load' => ['name' => 'inject', 'priority' => 0, 'events' => ['postLoad']],
-            'delete_on_update' => ['name' => 'clean', 'priority' => 50, 'events' => ['preRemove', 'postFlush']],
+            'delete_on_update' => ['name' => 'clean', 'priority' => 50, 'events' => ['preUpdate']],
```
garak, line 199: why?

garak (changes requested):

Tofandel:
Can we add a test to the extension to check that the listeners are registered?

garak (approved):

PR #1397, opened 2023-07-19
Update example with Embedded

qdequippe:
Add file property for Doctrine Embedded example to show how to map fields

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -271,6 +271,10 @@ class Product
 
     // ... other fields
 
+    // NOTE: This is not a mapped field of entity metadata, just a simple property.
+    #[Vich\UploadableField(mapping: 'products', fileNameProperty: 'image.name', size: 'image.size')]
+    private ?File $imageFile = null;
+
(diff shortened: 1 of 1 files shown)


docs/usage.md
```diff
@@ -271,6 +271,10 @@ class Product
 
     // ... other fields
 
+    // NOTE: This is not a mapped field of entity metadata, just a simple property.
```
garak, line 274: What would be the purpose of this?

garak (changes requested):

docs/usage.md
qdequippe: Your question is about the comment? if yes it's the same as in the [doc](https://github.com/dustin10/VichUploaderBundle/blob/master/docs/usage.md?plain=1#L96) without using the Embedded.

If it's about the PR globally it's to show mapping example for Embedded, for example `fileNameProperty: 'image.name', size: 'image.size'`

garak (approved):

Issue #625, opened 2016-09-06
[2.0] Roadmap for 2.0


Koc:
- [ ] Drop Symfony's support older than 2.7 or even 2.8
- [ ] Use options resolver for namers configuration
- [ ] Bump Doctrine version (requires for #618)
- [ ] Isolate Twig helper from templating component and make templating optional (see https://github.com/symfony/symfony/pull/13354)
- [ ] Maybe change configuration - set classes inside mapping, do not use `Uploadable` annotation (requires for #618)
- [ ] Add styleci

garak:
Doctrine version bumped in a8a9d25a05edf32a0c935e7f2ae3f5de8dd1ddf1

WolfgangDrescher:
I searched for `Symfony\Component\Templating` in this repo. If im correct the only place where VichUploaderBundle is using the `symfony/templating` dependency is in:

https://github.com/dustin10/VichUploaderBundle/blob/014f63c6ccf4856af134c8ddecf8e554771050eb/Templating/Helper/UploaderHelper.php#L13

But this UploaderHelper never using any methods of `Symfony\Component\Templating\Helper\Helper` (getCharset/setCharset). So if I'm correct we could just drop this class from extending the symfony templating helper dependency and remove it from composer.json.

garak:
@WolfgangDrescher extension is needed because UploaderHelper is tagged as `templating.helper` in container (so, it's needed to give user the opportunity of using such helper in pure-PHP templates).
Rather than remove extension, we should drop UploaderHelper entirely (so, removing support for PHP templates, following same path of Symfony) and move functionality to Twig extension

Erdou:
Hello @Koc ! Why "_Make final all classes that implement an interface_"? As it prevents to create any customized child Form Type 🤔

garak:
> Hello @Koc ! Why "_Make final all classes that implement an interface_"? As it prevents to create any customized child Form Type 🤔

Which form type?
Generally speaking, you won't extend a class implementing an interface: just use your own class

Erdou:
> 
> Which form type? Generally speaking, you won't extend a class implementing an interface: just use your own class

@Koc Supposing you're advocating for composition over inheritance, I noted Symfony's form types aren't final, indicating designed extendability. For example, I used that approach for the following AppImageType, aligning with some Symfony practices:


```
class AppImageType extends VichImageType
{
    protected int $maxImageSize;

[...]

PR #1406, opened 2023-11-05
Refactor: Make VichImageType class extensible by removing 'final'

Erdou:
This commit updates the VichImageType class by removing the 'final' keyword, thereby allowing for extension of the class. This modification maintains consistency with Symfony's practice of non-final form types, enabling developers to extend this form type to add custom functionality or modifications as needed.

### DIFF
--- a/src/Form/Type/VichImageType.php
+++ b/src/Form/Type/VichImageType.php
@@ -44,7 +44,7 @@ class VichImageType extends VichFileType
             'imagine_pattern' => null,
             'storage_resolve_method' => self::STORAGE_RESOLVE_URI,
             'attr' => [
-                'accept' => 'image/png,image/jpeg,image/gif,image/webp',
+                'accept' => 'image/*',
             ],
         ]);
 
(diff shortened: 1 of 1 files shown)


garak (changes requested):
Could we add the "accept" option as well?

Erdou:
@garak I'm not sure about what you are referring to. Can you precise? :)

garak:
The "attr" option in this comment https://github.com/dustin10/VichUploaderBundle/issues/625#issuecomment-1793430910

Erdou:
@garak Done 👍

src/Form/Type/VichImageType.php
```diff
@@ -43,6 +43,9 @@ public function configureOptions(OptionsResolver $resolver): void
             'image_uri' => true,
             'imagine_pattern' => null,
             'storage_resolve_method' => self::STORAGE_RESOLVE_URI,
+            'attr' => [
+                'accept' => 'image/png,image/jpeg,image/gif,image/webp',
```
garak, line 47: `'image/*'` is probably better (more synthetic and open to possible other formats)

garak (changes requested):

src/Form/Type/VichImageType.php
Erdou: > `'image/*'` is probably better (more synthetic and open to possible other formats)

Agreed, and done

garak (approved):

PR #1421, opened 2023-12-22
Fix Type: getUploadDir could not be null

RobinDev:


garak (changes requested):
I don't get the reason for this change. Even if there's one, CI is failing

RobinDev:
Reason : avoid to have to ignore the static analysis error or to do a useless string cast.

Normally, I fix the CI and if you think this pr is useless, it's okay for closing it.

garak:
bump

garak:
I'm sorry but I have to close this, it's too old and has no feedback.

PR #1428, opened 2024-02-07
update condition for object

mareckigit:
If form is used with Doctrine entity in Sonata Admin, `$form->getParent()->getData()` returns empty entity while you use create a new.

Probably VichFileType requires similar changes.

### DIFF
--- a/src/Form/Type/VichImageType.php
+++ b/src/Form/Type/VichImageType.php
@@ -71,7 +71,7 @@ class VichImageType extends VichFileType
         $view->vars['image_uri'] = null;
         $view->vars['download_uri'] = null;
 
-        if (null !== $object) {
+        if ((null !== $object && !is_object($object)) || (is_object($object) && $object->getId())) {
             if ($options['imagine_pattern']) {
                 if (null === $this->cacheManager) {
                     throw new \RuntimeException('LiipImagineBundle must be installed and configured for using "imagine_pattern" option.');


src/Form/Type/VichImageType.php
```diff
@@ -71,7 +71,7 @@ public function buildView(FormView $view, FormInterface $form, array $options):
         $view->vars['image_uri'] = null;
         $view->vars['download_uri'] = null;
 
-        if (null !== $object) {
+        if ((null !== $object && !is_object($object)) || (is_object($object) && $object->getId())) {
```
garak, line 74: I don't understand the purpose of this check.
The `$object` must be an object, otherwise the call to `resolveUriOption` would fail.
Moreover, we can't assume that a generic object has a `getId` method.

garak (changes requested):

src/Form/Type/VichImageType.php
mareckigit: More context:

- I have the form to create a new entity `Article`;
- `filePath` is property of related entity `private ?File $sourceFile = null;` and `#[ORM\ManyToOne(cascade: ['persist'])]` is set in `Article` for this relation;
- in `#[Vich\UploadableField` `fileNameProperty: 'sourceFile.filePath'` is set to inject data to `File` entity;
- VichImageType throws exception `PropertyAccessor requires a graph of objects or arrays to operate on, but it found type "NULL" while trying to traverse path "sourceFile.filePath" at property "filePath"`;
- `File` entity is optional in `Article` entity so I don't want to initialize it by `new File()` because cascade persist is used;
- VichImageType executes `resolveUriOption` on object which doesn't exist yet, so it doesn't make sens;
-  `$form->getParent()->getData()` returns `Article` not `File`;

Using `VichFileType` I can prevent this error by `'download_uri' => false` in form for new entity. There is no condition `if ($options['download_uri'] && $object) {` in `VichImageType` .

I can prepare next PR to add this missing condition to VichImageType, but if you expects here `$form->getParent()->getData()` an entity that has property for filePath, it dosen't work if related entity is used.

garak: So define your type class and use that instead of VichImageType.

mareckigit: Better idea is to update VichImageType to support  'download_uri' => false in the same way like in VicuFileType. Now even if is false, it calls resolveUriOption.

PR #1441, opened 2024-04-07
Add missing resolveUri method to FlysystemStorage

sovetski:
It resolves a lot of problems related to Flysystem.

Without this PR, if we are using S3, Cloudflare images or other cloud storage, the public url is not the CDN one but a local url which is incorrect.

For the people who will suggest using `uri_prefix`, as I mentionned here: https://github.com/dustin10/VichUploaderBundle/issues/1434#issuecomment-2041172155
In my case it is impossible to add `uri_prefix` manually, because in Cloudflare, the file name is not at the end, it is something like `https://imagedelivery.net/qsdazd54azd98a4z69d5/image3.jpg/public` the `/public` at the end makes it impossible because it is not just a prefix, we have to do some sprintf etc.

The problem existe since more than 8 years, I think it is time to automatise it with this PR.

I hope that it will be ok for maintainers.

Related:

https://github.com/dustin10/VichUploaderBundle/issues/1418
https://github.com/dustin10/VichUploaderBundle/issues/1434
https://github.com/dustin10/VichUploaderBundle/issues/502
https://github.com/dustin10/VichUploaderBundle/issues/1089
https://github.com/dustin10/VichUploaderBundle/issues/539

PR #1444, opened 2024-04-19
Revert changes of #1441

sovetski:
It looks like it is a breaking change, as mentionned by 3 people here: https://github.com/dustin10/VichUploaderBundle/pull/1441

Maybe in the future it will be implemented in a major release (I hope)

### DIFF
--- a/docs/storage/flysystem.md
+++ b/docs/storage/flysystem.md
@@ -57,7 +57,9 @@ vich_uploader:
 ### Issues with adapters public url
 
 If you are using certain adapters like S3, Cloudflare, etc., the generated public URL for assets may be incorrect.
-To resolve this, you can create your own [custom storage](https://github.com/dustin10/VichUploaderBundle/blob/master/docs/storage/custom.md) by extending the existing FlysystemStorage and adding this function:
+To resolve this, you can create your own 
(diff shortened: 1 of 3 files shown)


garak (changes requested):
What about updating the documentation to suggest this possible solution? So other devs can add this class (the old one I mean) and exploit this implementation.

sovetski:
@garak it's a good idea. I can add this information to the documentation for people who will want this feature. With custom storage, it will be simple to implement. Let me try to explain it.

garak (changes requested):
Sorry, yesterday I was in a hurry and didn't notice the invalid markdown. Could you fix it please?

sovetski:
@garak was it related to the last link that I added? If yes, I fixed it, and used a markdown validator to be sure that it is valid.

garak:
> @garak was it related to the last link that I added? If yes, I fixed it, and used a markdown validator to be sure that it is valid.

As you can see in the related Action, we enforce a custom rule about the maximum length of the lines.

sovetski:
> FlysystemStorage is final, you cannot extend it.

Yes, maybe my wording is not correct, but I was meaning that we have to duplicate its code

garak:
@sovetski still waiting for a fix in the markdown

sovetski:
@garak sorry for the delay, it looks OK now

PR #1446, opened 2024-04-22
Fix issue 144

wizhippo:
this keeps 2.x compatibility


garak:
Thanks for your proposal, but in the end we decided to revert the breaking change.

Issue #1449, opened 2024-05-08
Logging of storage operations


tarjei:
### Feature Request

If storage fails, then the errors should be logged.

For example in the FlySystemStorage module we got:

```

    protected function doRemove(PropertyMapping $mapping, ?string $dir, string $name): ?bool
    {
        $fs = $this->getFilesystem($mapping);
        $path = !empty($dir) ? $dir.'/'.$name : $name;

        try {
            $fs->delete($path);

            return true;
        } catch (FilesystemException $e) {
            return false;
        }
    }
```

This error is not propagated anywhere. It would be very usefull if it at least got logged.

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    |  no

#### Summary

Log errors in storagehandlers.

garak:
You can easily add a listener for the exception and perform whatever you want (logging, etc.)

tarjei:
@garak how? It seems to me it is just swallowed... :)

garak:
Sorry, I misread, you're right.
I think the method should throw the same `CannotWriteFileException` thrown by the `doUpload` method.
By the way, the previous exception should be passed.

tarjei:
@garak no problem :) This would require a try catch in UploadHandler wouldn't it? So that the exception is caught there and an event emitted? If the exception is thrown all the way then that would be a BC break - no?

garak:
On second thought, it's better to keep it like it is now.
The handler expects an exception for the upload and doesn't expect it from the remove.

Maybe you can exploit a combination of pre-remove and post-remove events to reach your goal.
Otherwise, redefining your own storage is always a possibility.

tarjei:
I would like to politely disagree. Debugging config errors etc in a storage handler is impossible the way it is setup now with silent fail.

The simplest solution would be a PR where the StorageHandlers get a logger and log each execption. This surfaces the error but does not change the overall behavior of the system. IMHO the other option is to let the UploadHandler catch the exception and emit an event.

PR #1450, opened 2024-05-10
If the underlying storage fails, emit an event

tarjei:
If the underlying storage fails, we emit an event that can be handled as the user wants.

The behaviour is to throw exceptions on write but not on remove as this seems to be the semantics as they are in the library prior to this change.

It seems to me that the error behavior of remove operations is a bit uneven. If a write fails in the native filesystem it seems to me that that will trigger some kind of error.

Anyhow, I think this keeps BC but introduces an event.

Ref #1449

### DIFF
--- a/src/Event/ErrorEvent.php
+++ b/src/Event/ErrorEvent.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Vich\UploaderBundle\Event;
+
+use Vich\UploaderBundle\Mapping\PropertyMapping;
+
+class ErrorEvent extends Event
+{
+    private \Throwable $throwable;
+
+    public function __construct(object $object, PropertyMapping $mapping, \Throwable $throwable)
+    {
+        parent::__construct($object, $mapping);
+        $this->throwable = $throwable;
+    }
+
+    public function getThrowable(): \Throwable
+    {
+        return $this->throwable;
+    }
+}
--- a/src/Event/Events.php
+++ b/src/Event/Events.php
@@ -54,4 +54,14 @@ final class Events
      * @Event("Vich\UploaderBundle\Event\Event")
      */
     public const POST_REMOVE = 'vich_uploader.post_remove';
+
+    /**
+     * Triggered if writing to storage fails.
+     */
+    public const UPLOAD_ERROR = 'vich_uploader.upload_error';
+
+    /**
+     * Triggered if removing the file from storage fails.
+     */
+    public const REMOVE_ERROR = 'vich_uploader.remove_error';
 }
--- a/src/Handler/UploadHandler.php
+++ b/src/Handler/UploadHandler.php
@@ -4,6 +4,7 @@ namespace Vich\UploaderBundle\Handler;
 
 use Symfony\Component\HttpFoundation\File\UploadedFile;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
+use Vich\UploaderBundle\Event\ErrorEvent;
 use Vich\UploaderBundle\Event\Event;
 use Vich\UploaderBundle\Event\Events;
 use Vich\UploaderBundle\FileAbstraction\ReplacingFile;
@@ -46,8 +47,12 @@ final class UploadHandler extends AbstractHandler
         }
 
         $this->dispatch(Events::PRE_UPLOAD, new Event($obj, $mapping));
-
-        $this->storage->upload($obj, $mapping);
+        try {
+            $this->storage->upload($obj, $mapping);
+        } catch (\Throwable $exception) {
+            $this->dispatch(Events::UPLOAD_ERROR, new ErrorEvent($obj, $mapping, $exception));
+            throw $exception;
+        }
         $this->injector->injectFile($obj, $mapping);
 
         $this->dispatch(Events::POST_UPLOAD, new Event($obj, $mapping));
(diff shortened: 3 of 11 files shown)


src/Event/ErrorEvent.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace Vich\UploaderBundle\Event;
+
+
+use Vich\UploaderBundle\Mapping\PropertyMapping;
+
+/**
```
garak, line 8: please remove this useless comment. Also, some style fixes are needed (see the actions)

src/Handler/UploadHandler.php
```diff
@@ -93,8 +99,13 @@ public function remove(object $obj, string $fieldName): void
         if ($preEvent->isCanceled()) {
             return;
         }
-
-        $this->storage->remove($obj, $mapping);
+        try {
+            $this->storage->remove($obj, $mapping);
+        } catch (CannotWriteFileException $cannotWriteFileException) {
+            throw $cannotWriteFileException;
```
garak, line 105: what's the point of throwing the exception and catching it below? Can't we just catch it directly?

src/Storage/StorageInterface.php
```diff
@@ -23,6 +23,7 @@ public function upload(object $obj, PropertyMapping $mapping): void;
      *
      * @param object          $obj     The object
      * @param PropertyMapping $mapping The mapping representing the field to remove
+     * @throw \Exception      Throws an exception
```
garak, line 26: An exception throwing in FileSystemStorage is still missing

garak (changes requested):

src/Storage/StorageInterface.php
tarjei: I tried to not change the behaviour as it is today. That's why I didn't add any.

src/Handler/UploadHandler.php
tarjei: This exception is not cought today. Again, just trying to follow existing behaviour - but I'm quite fine with catching it.

src/Event/ErrorEvent.php
tarjei: Which action by the way ? I tried to find it.

tarjei:
@garak thanks for a very quick review!

I'm still a bit unsure if an event is the correct approach. Maybe we could have a config option to either throw on failure or swallow it?

src/Storage/StorageInterface.php
garak: > I tried to not change the behaviour as it is today. That's why I didn't add any.

It should be caught too, and dispatch the new event.

src/Event/ErrorEvent.php
garak: https://github.com/dustin10/VichUploaderBundle/pull/1450/checks

tarjei: Thanks! Fixed now :) It seems to me that the PHPStan one is not related to this PR

garak: Not yet.

By the way, I see some English errors in your text, please check it with some language tool (something like Grammarly)

tarjei: My bad. I focused on the markdown issues.

tarjei: I also tried to improve the language as per your input :)

tests/Storage/GaufretteStorageTest.php
```diff
@@ -172,6 +172,8 @@ public function testThatRemoveMethodDoesDeleteFile(): void
      */
     public function testRemoveNotFoundFile(): void
     {
+        // the exception is cought in the UploadHandler.
```
garak, line 175: "caught"

src/Storage/FlysystemStorage.php
```diff
@@ -56,13 +56,8 @@ protected function doRemove(PropertyMapping $mapping, ?string $dir, string $name
...
 
-        try {
-            $fs->delete($path);
-
-            return true;
-        } catch (FilesystemException) {
-            return false;
-        }
+        $fs->delete($path);
+        return true;
```
garak, line 60: missing blank line before return

src/Storage/FileSystemStorage.php
```diff
@@ -41,11 +41,22 @@ protected function doRemove(PropertyMapping $mapping, ?string $dir, string $name
     {
         $file = $this->doResolvePath($mapping, $dir, $name);
 
-        return \file_exists($file) && \unlink($file);
+        if (!\file_exists($file)) {
+            throw new \Exception("File does not exist: ".$file);
+        }
+        if (!\unlink($file)) {
```
garak, line 47: We don't need separate exceptions here

src/Event/Events.php
```diff
@@ -53,5 +53,13 @@ final class Events
      *
      * @Event("Vich\UploaderBundle\Event\Event")
      */
-    public const POST_REMOVE = 'vich_uploader.post_remove';
+    public const POST_REMOVE  = 'vich_uploader.post_remove';
+    /**
```
garak, line 57: spacing and comments inconsistent with the rest of the file

src/Storage/FileSystemStorage.php
tarjei: OK, my rationale for doing so is that it provides a better debugging experience. I'll merge the two.

tarjei: (I plan to merge the commits when you're happy btw)

garak: This is linked below, in case you missed it: https://github.styleci.io/analyses/DynDKg?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

tarjei: Thanks! I didn't see that link no. I just applied the patch.

garak (approved):

garak:
PR is fine (already approved). Anyway, I meanwhile fixed the phpstan failure and would like to apply the check here.
Please rebase/merge master

tarjei:
@garak  done! :) Oh the bliss of all green

garak:
Thank you!

PR #1451, opened 2024-05-17
Optionally resolve the uri through flysystem

dheineman:
Based on the work of @sovetski in #1441 the PR will allow you to optionally have FlysystemStorage return the public URL of the filesystem instead. Doing it like this makes the new behaviour become opt-in and it does not require a new major version to release.

To enable simply add a configuration variable:
```diff
vich_uploader:
    db_driver: orm
    metadata:
        type: attribute
    storage: 'flysystem'
+   use_flysystem_to_resolve_uri: true
```


garak (changes requested):
Please rebase or merge master, the phpstan was fixed there

garak (approved):

PR #1453, opened 2024-05-24
Update vich_file_type.md

tomasnorre:
Isn't this expected to be `VichFileType::class` too?


garak (approved):
Good catch! Thank you

PR #1462, opened 2024-08-24
Ability to remove file for ReplacingFile

iksmail:
When I download a file from some source and want to add it to an entity, I use ReplacingFile and it creates the file according to the naming rules, but the downloaded file remains as well. I would like to be able to delete the downloaded file if needed.

### DIFF
--- a/src/FileAbstraction/ReplacingFile.php
+++ b/src/FileAbstraction/ReplacingFile.php
@@ -13,12 +13,18 @@ use Symfony\Component\HttpFoundation\File\File;
 class ReplacingFile extends File
 {
     protected bool $removeReplacedFile;
+    protected bool $removeReplacedFileOnError;
 
-    public function __construct(string $path, bool $checkPath = true, bool $removeReplacedFile = false)
-    {
+    public function __construct(
+        string $path,
+        bool $checkPath = true,
+        bool $removeReplacedFile = false,
+        bool $removeReplacedFileOnError = false
+    ) {
         parent::__construct($path, $checkPath);
 
         $this->removeReplacedFile = $removeReplacedFile;
+        $this->removeReplacedFileOnError = $removeReplacedFileOnError;
     }
 
     public function getClientOriginalName(): string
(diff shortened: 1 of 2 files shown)


src/FileAbstraction/ReplacingFile.php
```diff
@@ -5,16 +5,54 @@
...
     {
         return $this->getFilename();
     }
+
+    public function isRemoveReplacedFile(): bool
+    {
+        return $this->removeReplacedFile;
+    }
+
+    public function setRemoveReplacedFile(bool $removeReplacedFile): self
```
garak, line 40: I don't see any use for this setter

src/FileAbstraction/ReplacingFile.php
```diff
@@ -5,16 +5,54 @@
...
+
+        return $this;
+    }
+
+    public function isRemoveReplacedFileOnError(): bool
+    {
+        return $this->removeReplacedFileOnError;
+    }
+
+    public function setRemoveReplacedFileOnError(bool $removeReplacedFileOnError): self
```
garak, line 52: Same for this setter, it seems to be useless

garak (changes requested):

src/FileAbstraction/ReplacingFile.php
iksmail: You can use parameter passing in the constructor, or you can call this setter.
```
new ReplacingFile($file, true, true, true)
```
or
```
$replacingFile = new ReplacingFile($file);
$replacingFile->setRemoveReplacedFile(true);
```
or
```
(new ReplacingFile($file))->setRemoveReplacedFile(true);
```
Whichever is more convenient

src/FileAbstraction/ReplacingFile.php
iksmail: Similar to above

src/FileAbstraction/ReplacingFile.php
garak: Still, I don't see a case where calling the setter could be convenient. If the only purpose is offering a setter alternative to the constructor, it's a weak purpose. It's better to keep the object immutable.

iksmail: Removed the setters

PR #1480, opened 2024-11-29
Use explicit nullables types

aleho:
Fixes a deprecation warning in preparation for PHP 8.4.


aleho:
Wow. You're faster than I can reload a tab. ❤️

PR #1486, opened 2025-01-20
Fix tests

guillaume-sainthillier:
This PR fix some tests that were skipped because of bad "@requires" annotation

### DIFF
--- a/tests/Storage/Flysystem/AbstractFlysystemStorageTestCase.php
+++ b/tests/Storage/Flysystem/AbstractFlysystemStorageTestCase.php
@@ -78,7 +78,7 @@ abstract class AbstractFlysystemStorageTestCase extends StorageTestCase
 
         $this->filesystem
             ->expects(self::once())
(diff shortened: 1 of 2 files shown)


tests/Storage/Flysystem/AbstractFlysystemStorageTestCase.php
```diff
@@ -157,9 +152,9 @@ public static function pathProvider(): array
         return [
             //     dir,   path,                     relative
             [null,  'file.txt',               true],
-            [null,  '/absolute/file.txt',     false],
+            [null,  'file.txt',               false],
```
garak, line 155: any reason for changing these values?

garak (changes requested):

garak (approved):

garak:
The tests are currently failing because 
> Trying to configure method "publicUrl" which cannot be configured because it does not exist, has not been specified, is final, or is static

PR #1489, opened 2025-01-20
update objects and methods typehint

guillaume-sainthillier:


src/Naming/Base64Namer.php
```diff
@@ -17,7 +17,7 @@ class Base64Namer implements NamerInterface, ConfigurableInterface
     protected const ALPHABET = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ-_';
 
     /** @var int Length of the resulting name. 10 can be decoded to a 64-bit integer. */
-    protected $length = 10;
```
garak, line 20: This is a protected property of an extensible class, altering it is a BC break. If you want this to be merged soon, you need to revert this change.

tests/Fixtures/TestBundle/src/Entity/Image.php
```diff
@@ -48,8 +35,6 @@ public function getId(): ?int
      * bundle's configuration parameter 'inject_on_load' is set to 'true' this setter
      * must be able to accept an instance of 'File' as the bundle will inject one here
      * during Doctrine hydration.
-     *
-     * @param File|\Symfony\Component\HttpFoundation\File\UploadedFile|null $image
      */
     public function setImageFile(File $image = null): void
```
garak, line 39: I don't know why this is not `?File`, do you mind fixing it?

garak (changes requested):

src/Naming/Base64Namer.php
guillaume-sainthillier: @garak fixed

garak (approved):

PR #1491, opened 2025-01-20
Improve public uri

guillaume-sainthillier:
This PR needs #1486 before to be merged.

When using publicUri with flysystem, it could be necessary to have more context (the object, mapping and fieldName) to generate a public uri

### DIFF
--- a/src/Storage/FlysystemStorage.php
+++ b/src/Storage/FlysystemStorage.php
@@ -97,7 +97,12 @@ final class FlysystemStorage extends AbstractStorage
         $fs = $this->getFilesystem($mapping);
 
         try {
-            return $fs->publicUrl($path);
+            return $fs->publicUrl($path, [
+                'object' => $obj,
+                'fieldName' => $fieldName,
+                'className' => $className,
+                'mapping' => $mapping,
+            ]);
         } catch (FilesystemException|UndefinedMethodError) {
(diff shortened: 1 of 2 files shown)


src/Storage/FlysystemStorage.php
```diff
@@ -97,7 +97,12 @@ public function resolveUri(object|array $obj, ?string $fieldName = null, ?string
         $fs = $this->getFilesystem($mapping);
 
         try {
-            return $fs->publicUrl($path);
+            return $fs->publicUrl($path, [
```
garak, line 100: Could you elaborate a bit on this? I don't get why we need to pass our internal information (like the mapping and the field name) to Flysystem

garak:
You can ignore the phpstan problem, but the tests need to be fixed

src/Storage/FlysystemStorage.php
guillaume-sainthillier: @garak sometimes, the url generator must be aware of which object and which fields is getting generated.

For example, an `User` entity with `profilePictureFile` (public) and `bankDetailFile` (private) fields.
The generator must generate a public url for one, and a private url with authorization for the other field.

Passing internals like mapping would not be needed in this case so it can be removed if you think it's not worth it

garak (approved):

PR #1505, opened 2025-04-17
use attributes in commands

### LINKED ISSUE #1502: getDefaultName deprecated in commands
> Since symfony/console 7.3: Overriding "Command::getDefaultName()" in "Vich\UploaderBundle\Command\MappingDebugClassCommand" is deprecated and will be removed in Symfony 8.0, use the #[AsCommand] attribute instead.

Since the attribute is supported in every Symfony version this bundle uses, let's remove the method and use the attribute instead

garak:
Fix #1502

### DIFF
--- a/src/Command/MappingDebugClassCommand.php
+++ b/src/Command/MappingDebugClassCommand.php
@@ -2,6 +2,7 @@
 
 namespace Vich\UploaderBundle\Command;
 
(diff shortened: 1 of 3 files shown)


laurentmuller:
Maybe, it will be good to put also the description within the atttribute:

```php
#[AsCommand(name: 'vich:mapping:debug-class', description: 'Debug a class.'')]
```

PR #1508, opened 2025-05-27
Allow symfony 5.4 symfony/event-dispatcher-contracts

wizhippo:
Symfony 5.4 is not compatible with symfony/event-dispatcher-contracts 3.1 and 2.2 should be allowed.

I do not see any > 2.2 specific features used in VichUploaderBundle

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "symfony/config": "^5.4 || ^6.0 || ^7.0",
         "symfony/console": "^5.4 || ^6.0 || ^7.0",
         "symfony/dependency-injection": "^5.4 || ^6.0 || ^7.0",
-        "symfony/event-dispatcher-contracts": "^3.1",
+        "symfony/event-dispatcher-contracts": "^2.2 || ^3.1",
         "symfony/http-foundation": "^5.4 || ^6.0 || ^7.0",
         "symfony/http-kernel": "^5.4 || ^6.0 || ^7.0",
         "symfony/mime": "^5.4 || ^6.0 || ^7.0",


garak:
Could you clarify how the incompatibility applies?

wizhippo:
I have an older component that uses symfony 5.4 and relies on symfony/event-dispatcher-contracts ^2.2 which would be accurate for 5.4 as it requires "symfony/contracts": "^2.1",    symfony/contracts has replace "symfony/event-dispatcher-contracts": "self.version" which would resolve to symfony/event-dispatcher-contracts 2.x so 3.1 would fail.

because vich has it fixed at ^3.1 i get a conflict

garak:
It looks like the incompatibility is in your project, not in Symfony.
By the way, we have tests in place assuring that this bundle works even with Symfony 5.4.0

wizhippo:
Symfony 5.4 is compatible with both symfony/event-dispatcher-contracts:^2.2 and ^3.1 — the latter due to forward compatibility. However, the version that was originally released and with Symfony 5.4 was ^2.2. Third-party bundles (such as ibexa/core-search) still rely on ^2.2, which is entirely valid within the Symfony 5.4 ecosystem.

By requiring ^3.1 exclusively, VichUploaderBundle introduces a constraint that breaks compatibility with other Symfony 5.4-compatible packages.

While your tests may pass with ^3.1, they do not account for the full compatibility matrix of Symfony 5.4, particularly with older or stable bundles that haven't moved to ^3.1. This creates a situation where VichUploaderBundle is technically not fully compatible with Symfony 5.4.

I get you would rather the other package update it's requirements to more modern standards, but the issue really is the constraint in VichUploaderBundle.

PR #1520, opened 2025-07-29
Update usage.md

kstefanini:
Hey!
I just ran into difficulties to make things work so I propose to move information to a more readable place

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -62,6 +62,11 @@ Next, you have to create the two fields needed for the bundle to work:
      object after the form is submitted. This should *not* be persisted to the
      database, but you *do* need to annotate it.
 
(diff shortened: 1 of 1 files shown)


garak:
There's already an "important" block below.
Instead of adding a redundant note, we should enhance that with the proper markdown, e.g. `[!IMPORTANT]` instead of `**Important:**`.
By the way, all the blocks should be enhanced in the same way

kstefanini:
Thanks for your quick reply, I've updated my proposal

garak:
At this point, why not update all the other pages as well?

garak:
If you rebase, the CS failures should go.
After that, if you could please apply the suggested changes, I'll be glad to merge this

PR #1524, opened 2025-08-13
⬆️  allow Symfony 8

### LINKED ISSUE #1519: Symfony 8 support
We know that Symfony 8.0 is currently still in its development phase, anyway it would be nice to support it, allowing developers to test it in real projects before its stable release.

garak:
Fix #1519


garak:
The failing tests are unrelated.
See https://github.com/sebastianbergmann/phpunit/issues/6226#issuecomment-2939783024 for an explanation

PR #1529, opened 2025-08-23
do not inject container in PropertyMappingResolver

### LINKED ISSUE #1448: Autowire services that implement NamerInterface
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| BC Break    | no

#### Summary

Instead or requiring namer services to be manually registered.

```yaml
# services.yaml
    App\Service\ProjectImageNamerService:
        public: true
````

The compiler pass could look for services that implement NamerInterface and make them public.  One less step when creating a custom namer service.

garak:
Fix #1448

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,4 +1,8 @@
-# Upgrading from v2.5 to v2.6
+# Upgrading from v2.7 to v2.8
+
+* Namers are not public anymore. If you uses a custom namer, you can now make it private.
+
+# Upgrading from v2.6 to v2.7
 
 * Now the original extension '.xlsb' is retained even if the mime type is guessed as 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'.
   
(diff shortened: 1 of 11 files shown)


Dallas62:
Well, it's not a bad upgrade, but a breaking one for those who follow the documentation:
https://github.com/dustin10/VichUploaderBundle/blob/master/docs/directory_namer/howto/create_a_custom_directory_namer.md

When using the service name, these change break the initial behaviour.

Leading to:
UnexpectedValueException: Directory namer service "directory_namer" not found.

Need to update the documentation to use Class Names, or restore previous implementation by allowing to get the service by name.

garak:
> Well, it's not a bad upgrade, but a breaking one f
> Leading to: UnexpectedValueException: Directory namer service "directory_namer" not found.

Nice catch. We should be able to fix it by using the default-index-method option in the definition of the tagged iterafor. PR welcome

garak:
@Dallas62 please check if PR #1537 solves the problem for you

PR #1537, opened 2025-09-11
Fix namer resolving

### LINKED ISSUE #1536: Namer service not found
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.8.0

#### Summary

After upgrading to v2.8.0 I now encounter the issue where I now get a [UnexpectedValueException] "Namer service "servicename.mappingname" not found"

I've traced the issue to the fact that some of my mappings contain the "." character.
In the method getNamer in \Vich\UploaderBundle\Mapping\PropertyMappingResolver the service name is checked for the existence of the "." character which is needed as the namer is registered internally with
[servicename|Namer::class].[mappingname]

however getNamer uses strrpos to split the namers service name from the mapping name.
As my mapping contains one or more "." characters it returns the wrong location.

using strpos instead of strrpos in \Vich\UploaderBundle\Mapping\PropertyMappingResolver::getNamer 
solves the issue for me

I can however imagine that if the service name contains a "." character, which it might well do, this would break the reverse case.

So I would propose not using a "." as the delimiter for the service and the mapping when registering the namer services in the PropertyMappingResolver since both services and mapping names can contain a "."

#### Previous behavior

A mapping name can contain a "."

#### Current behavior

Mappings can no longer contain a "."
[UnexpectedValueException] "Namer service "servicename.mappingname" not found"

#### How to reproduce

Create a mapping containing a "." character

garak:
Fix #1536

### DIFF
--- a/config/factory.xml
+++ b/config/factory.xml
@@ -5,8 +5,8 @@
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
         <service id="vich_uploader.property_mapping_resolver" class="Vich\UploaderBundle\Mapping\PropertyMappingResolver" public="false">
-            <argument type="tagged_iterator" tag="vich_uploader.namer"/>
-            <argument type="tagged_iterator" tag="vich_uploader.dir_namer"/>
+            <argument type="tagged_iterator" tag="vich_uploader.namer" default-index-method="getId"/>
+            <argument type="tagged_iterator" tag="vich_uploader.dir_namer" default-index-method="getId"/>
             <argument>%vich_uploader.mappings%</argument>
             <argument>%vich_uploader.default_filename_attribute_suffix%</argument>
         </service>
--- a/src/Mapping/PropertyMappingResolver.php
+++ b/src/Mapping/PropertyMappingResolver.php
@@ -18,8 +18,8 @@ use Vich\UploaderBundle\Util\ClassUtils;
 final class PropertyMappingResolver implements PropertyMappingResolverInterface
 {
     /**
-     * @param iterable<NamerInterface>          $namers
-     * @param iterable<DirectoryNamerInterface> $dirNamers
+     * @param iterable<string, NamerInterface>          $namers
+     * @param iterable<string, DirectoryNamerInterface> $dirNamers
      */
     public function __construct(
         private readonly iterable $namers,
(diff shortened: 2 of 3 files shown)


Tofandel:
Hmm, this still doesn't solve the issue

I had the following

```php
    $services->alias('app.microid_namer.case_files', MicroIdNamer::class)
        ->public();

$config->mappings('case_files', [
        'uri_prefix' => '/case_uploads',
        'upload_destination' => 'oneup_flysystem.case_files_filesystem',
        'namer' => 'app.microid_namer',
        'directory_namer' => [
            'service' => 'vich_uploader.namer_directory_property',
            'options' => [
                'property' => 'directory_name',
                'transliterate' => false,
            ],
        ],
        'delete_on_remove' => true,
        'delete_on_update' => true,
    ]);
```

With this update I get 'Namer service "app.microid_namer.case_files" not found.'


I dumped $this->namers.id and it contains

```
Array ( [0] => App\MicroId\MicroIdNamer [1] => Vich\UploaderBundle\Naming\UniqidNamer [2] => Vich\UploaderBundle\Naming\PropertyNamer [3] => Vich\UploaderBundle\Naming\OrignameNamer [4] => Vich\UploaderBundle\Naming\HashNamer [5] => Vich\UploaderBundle\Naming\Base64Namer [6] => Vich\UploaderBundle\Naming\SmartUniqueNamer [7] => vich_uploader.namer_property )
```

So the aliases seem to not be present anymore causing the lookup to fail

If I use a class in the namer config I still have an issue with
Directory namer service "vich_uploader.namer_directory_property" not found.

`$this->dirNamers.id` contains

```
Array ( [0] => Vich\UploaderBundle\Naming\SubdirDirectoryNamer [1] => Vich\UploaderBundle\Naming\PropertyDirectoryNamer [2] => Vich\UploaderBundle\Naming\CurrentDateTimeDirectoryNamer [3] => Vich\UploaderBundle\Naming\ConfigurableDirectoryNamer )
```

So it seems the `index-by getId` doesn't resolve to the service name but to the class (Except for vich_uploader.namer_property for some reason)

So for now all the services need to be changed to classes 

```
'directory_namer' => [
            'service' => Vich\UploaderBundle\Naming\PropertyDirectoryNamer::class,
            'options' => [
                'property' => 'directory_name',
                'transliterate' => false,
            ],
        ],
```

PR #1538, opened 2025-09-19
Support multiple extensions for MIME types in FileExtensionTrait

endelwar:
Refactored the extension mapping to use arrays instead of single values, allowing multiple file extensions to share the same MIME type. Added support for srt/vtt subtitle files and kml geospatial data.

### DIFF
--- a/src/Naming/Polyfill/FileExtensionTrait.php
+++ b/src/Naming/Polyfill/FileExtensionTrait.php
@@ -10,9 +10,9 @@ trait FileExtensionTrait
 {
     // extensions safe to keep
     private static array $keep = [
-        'txt' => 'csv',
-        'xml' => 'gpx',
-        'xlsx' => 'xlsb',
+        'txt' => ['csv', 'srt', 'vtt'],
+        'xml' => ['gpx', 'kml'],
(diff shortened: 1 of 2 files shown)


garak (approved):
👍🏻

laurentmuller:
Just a point of view. 

When someone ask for a new MIME type exception, we have to update the trait and create a new version.

Maybe, it will be better to allow user to configure it and create a compiler pass to merge existing mapping within the configuration.

Yaml Example:

```yaml
vich_uploader:
    extensions:
        txt:  ['csv', 'srt', 'vtt']
        xml:  ['gpx', 'kml']
        xlsx: ['xlsb']
```

garak:
So end users will keep changes for themselves instead of sharing with us?

laurentmuller:
Yes, that's kind of the idea. You can keep the existing list and merge with the list defined in the configuration.

endelwar:
IMHO, it's better to share new extensions and MIME types with the community. For those who don't want to share, there is the `namer_keep_extension` parameter (#1540).

laurentmuller:
What if someone wants one of the defined extensions and not the others?

PR #1541, opened 2025-09-22
Add FileRequired validator constraint

endelwar:
Provides validation for file upload fields ensuring either existing file or new upload is present.
Extends Symfony NotBlank constraint with file-specific logic supporting UploadedFile, ReplacingFile, and existing Vich\UploaderBundle\Entity\File instances.


endelwar:
To make the tests pass on Symfony 5.4.* with the lowest dependencies, we need to set the minimum Symfony version to at least 5.4.21 (better 5.4.22), where ConstraintValidatorTestCase was made compatible with PHPUnit 10.

garak:
Is this PR ready? If not, what is still pending?

endelwar:
Just finished last cleanup and local tests: ready for review

nicodemuz:
It would be great to document also other common use cases, for example how to upload an optional file that is an image and with a given maximum file size.

docs/usage.md
```diff
@@ -271,6 +271,34 @@ All options are listed below:
...
+validation logic:
+
+```php
+use Vich\UploaderBundle\Validator\Constraints as VichAssert;
+
+class Product
+{
+    // ... other fields
+
+    #[Vich\UploadableField(mapping: 'products', fileNameProperty: 'image.name')]
```
nicodemuz, line 287: Is `fileNameProperty` a valid example here? Can a property name contain a dot? Does it reference some other Entity? Would be great to explain in more detail.

garak: The property is obtained using the PropertyAccess component. Youn can referer to the documentation at https://symfony.com/doc/current/components/property_access.html

PR #1542, opened 2025-09-23
Update PHPStan configuration and fix type issues

endelwar:
Modernizes PHPStan setup with dist configuration file and PHP-based baseline.
Fixes various type-related issues including template variance violations, array unpacking with named arguments, and return type inconsistencies in storage implementations.
Updates CI workflow to use new configuration structure.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -17,7 +17,7 @@ jobs:
               with:
                   php-version: 8.3
             - run: |
-                  composer require --no-update doctrine/annotations:"^2.0" liip/imagine-bundle:"^2.13" phpstan/phpstan:"^1.12" 
+                  composer require --no-update doctrine/annotations:"^2.0" liip/imagine-bundle:"^2.13" phpstan/phpstan:"^2" 
                   composer install --ignore-platform-reqs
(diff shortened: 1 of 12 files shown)


garak:
Great! Do you think there's room for improving the phpstan level to 7?

garak:
It seems that moving the configuration to PHP causes more harm than good.
I'm not convinced about renaming the configuration file either... What's the use case for a developer-specific configuration?

endelwar:
> Great! Do you think there's room for improving the phpstan level to 7?

There’s quite a lot of stricter type checking required to reach level 7 (especially in the mocked objects in tests), but I don’t think it’s impossible.

endelwar:
> What's the use case for a developer-specific configuration?

Using `phpstan.dist.neon` leave space for a local `phpstan.neon` where a developer can experimtn without touching project files

garak:
We could try to introduce phpstan-phpunit, but probably it's better to do it in a separate PR

endelwar:
Reverted phpstan php config to neon. It seems like `twig.exception_controller` is messing up tests on php8.4+Symfony7.4

PR #1554, opened 2025-10-23
Migrate configuration from XML to PHP

endelwar:
Follows the same approach as master branch (46fdbd8), preserving all 3.x-specific features (cleanup command and interface aliases).


garak:
I was thinking about this in the last days: do we need a PR for branch 3 for every PR already merged in master? Isn't cherry-picking enough?

endelwar:
Usually, cherry-picking would be enough, but in 3.x there were new interfaces and the new cleanup command, which required some adjustments in `Command.php` and `Factory.php` (the other PHP config files are the same as in master).

garak (approved):

PR #1586, opened 2026-05-28
fix namer services deprecations

### LINKED ISSUE #1585: Deprecation in tagged iterators
When running VichUploaderBundle v3.0.0-rc3 with Symfony 8.1+, the following deprecation notice is triggered during container compilation:

> Since symfony/dependency-injection 8.1: The $defaultIndexMethod and $defaultPriorityMethod arguments
> of tagged locators and iterators are deprecated, use the #[AsTaggedItem] attribute instead.

## Root Cause

In `config/factory.php`, the `PropertyMappingResolver` service is wired using `tagged_iterator()` with the now-deprecated `defaultIndexMethod` named argument:

```php
tagged_iterator('vich_uploader.namer', defaultIndexMethod: 'getId'),
tagged_iterator('vich_uploader.dir_namer', defaultIndexMethod: 'getId'),
```

## Expected Fix
Add `#[AsTaggedItem]` to the built-in namer and directory namer classes with the appropriate index value (the service alias / string used to look them up in PropertyMappingResolver), and remove the defaultIndexMethod argument from the two tagged_iterator() calls in config/factory.php.
For example:

```php
use Symfony\Component\DependencyInjection\Attribute\AsTaggedItem;

#[AsTaggedItem(index: 'vich_uploader.namer.uniqid')]
class UniqidNamer implements NamerInterface { ... }
```

And in config/factory.php:
```php
tagged_iterator('vich_uploader.namer'),
tagged_iterator('vich_uploader.dir_namer'),
```

## Notes
`PropertyMappingResolver::getNamer()` also matches by `$namer::class`, so numeric iteration may already be sufficient for the default namers, but explicit `#[AsTaggedItem]` indexes are needed to preserve lookup by service alias for any custom namers registered by users.

The fix must target the 3.x branch, but we can consider the possibility of backporting the fix to the 2.x branch

garak:
Fix #1585

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -6,6 +6,10 @@
 * The deprecated `AnnotationInterface` has been removed. Use `AttributeInterface` instead.
 * `AttributeReader` deprecated methods have been removed: use `getClassAttribute()` instead of `getClassAnnotation()`, `getPropertyAttribute()` instead of `getPropertyAnnotation()`.
 
+## Symfony 8.1+ compatibility note
+
+* If you configure a custom `namer` or `directory_namer` by service id (for example `my.custom_namer`) instead of by class name, index that service in tagged iterators using `#[AsTaggedItem(index: 'my.custom_namer')]` (or equivalent tag `index` config).
+
 # Upgrading from v2.8 to v2.9
 
 ## Deprecations
--- a/config/factory.php
+++ b/config/factory.php
@@ -12,8 +12,8 @@ return static function (ContainerConfigurator $container): void {
 
     $services->set('vich_uploader.property_mapping_resolver', PropertyMappingResolver::class)
         ->args([
-            tagged_iterator('vich_uploader.namer', defaultIndexMethod: 'getId'),
-            tagged_iterator('vich_uploader.dir_namer', defaultIndexMethod: 'getId'),
+            tagged_iterator('vich_uploader.namer'),
+            tagged_iterator('vich_uploader.dir_namer'),
             param('vich_uploader.mappings'),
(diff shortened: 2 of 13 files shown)


deluxetom (approved):

milanqtx (approved):

milanqtx:
Worth backporting to `2.x`. When multiple mappings reference the same `directory_namer` service with different `options`, they all share one singleton that `PropertyMappingResolver::resolve()` reconfigures on every call, so every mapping ends up observing the last resolved mapping's options.

Hit this on v2.9.4 with two `PropertyDirectoryNamer` mappings — one `transliterate: true`, one `transliterate: false`. The flag flipped depending on resolve order and paths like `1234/foo/1` got rewritten to `1234-foo-1`.

garak:
@milanqtx the 2.x branch won't support Symfony 9, so we don't care about deprecations there

endelwar:
> Worth backporting to `2.x`. When multiple mappings reference the same `directory_namer` service with different `options`, they all share one singleton that `PropertyMappingResolver::resolve()` reconfigures on every call, so every mapping ends up observing the last resolved mapping's options.
> 
> Hit this on v2.9.4 with two `PropertyDirectoryNamer` mappings — one `transliterate: true`, one `transliterate: false`. The flag flipped depending on resolve order and paths like `1234/foo/1` got rewritten to `1234-foo-1`.

I was bitten by this bug too, I prepared 2 PRs (2.x #1610 and 3.x #1611).

garak:
Superseded by #1613

PR #1595, opened 2026-06-01
Fix "null as array offset" deprecation in FileExtensionTrait

endelwar:
`guessExtension()` can return null, which reached `isset(self::$keep[null])`. Guard against it explicitly; behavior unchanged.

### DIFF
--- a/src/Naming/Polyfill/FileExtensionTrait.php
+++ b/src/Naming/Polyfill/FileExtensionTrait.php
@@ -50,7 +50,7 @@ trait FileExtensionTrait
             return $this->getOriginalExtension($file);
         }
 
-        if ('' !== ($extension = $file->guessExtension())) {
+        if (null !== ($extension = $file->guessExtension()) && '' !== $extension) {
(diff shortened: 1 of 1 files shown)


tests/bootstrap.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+use Symfony\Component\ErrorHandler\ErrorHandler;
+
+require \dirname(__DIR__).'/vendor/autoload.php';
+
+// see https://github.com/symfony/symfony/issues/53812#issuecomment-1962740145
```
garak: This issue was closed a while ago, and this bundle currently works without the "boostrap" workaround.

endelwar: I added commit bb21d40aa03ba2ea284ffb9b943f3e4cd4b38301 to suppress the exception-handler warning and reduce PHPUnit noise. Would you like me to remove it?

garak: It seems to bring no value anymore

garak (approved):

PR #1596, opened 2026-06-01
Modernize Docker test images & fix PHP 8.3 / 8.5

endelwar:
The `docker/Dockerfile{83,84,85}` images used by `make tests` were built with the
now-deprecated `pecl install`, shipped the developer's local `vendor/` into the
image, and forced an incompatible Symfony version on PHP 8.3.

### Changes
- **pecl → [docker-php-extension-installer](https://github.com/mlocati/docker-php-extension-installer)** for the `mongodb` extension and Composer. `pecl` is officially deprecated by the MongoDB PHP driver. Build deps are now auto-removed.
- **Unify `mongodb` on `2.3.3`** across all three images (was 1.17.0 / 1.21.0 / 2.1.2).
- **Pin 8.5 to stable `php:8.5-alpine`** (was `8.5.0rc3-alpine` which is unavailable) — the suite now runs on PHP 8.5.
- **Drop `--ignore-platform-reqs`**: on the PHP 8.3 image it let Composer install Symfony 8.x (which requires `php >=8.4.1`) instead of the highest compatible release, so the whole suite failed to load with a `ParseError` on Symfony's property-hook syntax. Without the flag, PHP 8.3 resolves Symfony 7.4 and the suite passes. `composer check-platform-reqs` is green on all three images, so the flag was only bypassing the PHP version constraint — no required extension is missing.
- **Add `.dockerignore`** so a `vendor/` built on the developer's machine is no longer copied into the Linux image.
- **BuildKit cache mount** for Composer's download cache (replaces `composer clear-cache`): fast rebuilds without freezing dependency resolution.

### Results

| Metric | PHP 8.3 | PHP 8.4 | PHP 8.5 |
|---|---|---|---|
| Image size before | 498 MB | 503 MB | 529 MB |
| Image size after | **190 MB** | **194 MB** | **221 MB** |
| Reduction | −62% | −61% | −58% |
| Tests | ✅ 356 | ✅ 356 | ✅ 356 (was broken) |

| Build metric | Before | After |
|---|---|---|
| Build context transferred | ~50 MB | **17.8 kB** |
| Composer re-download on rebuild | full | **0** (cache mount) |
| Composer step (warm cache) | 5.0 s | **2.4 s** |

All 356 tests pass on PHP 8.3, 8.4 and 8.5.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,11 @@
+# Keep the build context small and avoid shipping host-built artifacts.
+.git
+.github
+.idea
+.DS_Store
+.*.cache
+.phpunit.result.cache
+vendor/
+composer.lock
+coverage/
+docs/
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
+/.dockerfile*     export-ignore
 /.git*            export-ignore
 /.php-cs-fixer*   export-ignore
 /CONTRIBUTING.md  export-ignore
--- a/docker/Dockerfile83
+++ b/docker/Dockerfile83
@@ -1,43 +1,20 @@
 FROM php:8.3-alpine
 
-ARG MONGODB_VERSION=1.17.0
+ARG MONGODB_VERSION=2.3.3
 
-# Install mongo
-RUN set -eux; \
-	apk add --no-cache --virtual .build-deps \
-		$PHPIZE_DEPS \
-    	openssl-dev \
-	; \
-	pecl install \
-	    mongodb-${MONGODB_VERSION} \
-    ; \
-	docker-php-ext-enable \
-	    mongodb \
-	; \
-	pecl clear-cache; \
-	\
-	runDeps="$( \
-    scanelf --needed --nobanner --format '%n#p' --recursive /usr/local/lib/php/extensions \
-      | tr ',' '\n' \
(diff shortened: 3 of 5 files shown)


endelwar:
PHP-CS-Fixer issues not related to PR

garak (approved):

garak:
@endelwar I just tried to run `TARGET=84 make tests` on branch 3.x and got the following problem (omitting the previous steps for brevity):

>Step 7/8 : RUN --mount=type=cache,target=/tmp/composer-cache 	COMPOSER_CACHE_DIR=/tmp/composer-cache composer update --prefer-dist
the --mount option requires BuildKit. Refer to https://docs.docker.com/go/buildkit/ to learn how to build images with BuildKit enabled
make[1]: *** [Makefile:29: vichuploader-image] Errore 1
make[1]: uscita dalla directory «/foo/VichUploaderBundle»
make: *** [Makefile:7: tests] Errore 2

endelwar:
BuildKit is standard from Docker Engine 23 released in february 2023, which versione are your running?

garak:
$ docker --version
Docker version 29.1.3, build 29.1.3-0ubuntu4.1

Edit: solved with a `sudo apt install docker-buildx`

PR #1603, opened 2026-08-10
Fix unrelated attribute instantiation

### LINKED ISSUE #1442: The attribute reader should not trigger an exception for other attributes
### Bug Report

Jetbrains offers several attributes to help adding metadata to the code. 
For example,  #[Deprecated], #[ArrayShape] and others.
After adding some of these to the code an exception was thrown by this bundle's attribute reader. 
I believe this to be a bug. Attribute readers from Doctrine or Symfony don't have this problem.

Error message: _Attribute class "JetBrains\PhpStorm\Deprecated" not found_
Error FQCN:  \Error

|    Q            |   A
|---------------- | ------
| BC Break        | no
| Bundle version  | 2.3.1
| Symfony version | *
| PHP version     | >= 8

#### Summary

This bundle should not break when I use the attributes from jetbrains.

#### Current behavior

When I use the #[Deprecated] attribute for example,  the attribute reader in this bundle will throw an exception.
This happens on line 60 where the attribute class is instantiated. 

#### How to reproduce

Just add one of the attributes to a property. Possibly, the class must be an entity that that is using this bundle.
```

    #[ORM\Column(type: "string")]
    #[Deprecated()]
    protected ?string $helloWorld;
```

#### Expected behavior

Exception is not thrown.

#### Suggested solution

I would suggest to add a try catch block. I already verified that this change would fix the issue.

```
        foreach ($attributes as $attribute) {
            $attributeName = $attribute->getName();
            try {
                $instance = $attribute->newInstance();
            } catch(\Error $e) {
                 continue;
            }

            if (!$instance instanceof AnnotationInterface) {
                continue;
            }

            $instances[$attributeName] = $instance;
        }
```

ousamabenyounes:
## Summary

Ignore unrelated PHP attributes before instantiating upload mapping attributes.

The reader previously instantiated every reflected attribute and only then checked whether it implemented `AttributeInterface`. Unrelated attributes could therefore throw or trigger side effects while Vich metadata was read.

Fixes #1442.

## Test verification (RED → GREEN)

Tested with PHP 8.3 in the repository Docker environment.

RED on unmodified `origin/master` with only the regression test:

```text
LogicException
tests/Metadata/Driver/AttributeReaderTest.php:183
src/Metadata/Driver/AttributeReader.php:122
ERRORS!
Tests: 9, Assertions: 12, Errors: 1.
```

GREEN with the fix:

```text
OK (9 tests, 13 assertions)
```

Full local CI replay was iso-or-better than the upstream baseline: both runs retain the same pre-existing `GaufretteStorageTest` error; PHPStan and PHP-CS-Fixer pass. Changed executable-line coverage: `2/2` (100%).

### DIFF
--- a/src/Metadata/Driver/AttributeReader.php
+++ b/src/Metadata/Driver/AttributeReader.php
@@ -119,13 +119,12 @@ final class AttributeReader
 
         foreach ($attributes as $attribute) {
             $attributeName = $attribute->getName();
-            $instance = $attribute->newInstance();
 
-            if (!$instance instanceof AttributeInterface) {
+            if (!\is_a($attributeName, AttributeInterface::class, true)) {
                 continue;
             }
 
-            $instances[$attributeName] = $instance;
+            $instances[$attributeName] = $attribute->newInstance();
         }
 
         return $instances;
--- a/tests/Metadata/Driver/AttributeReaderTest.php
+++ b/tests/Metadata/Driver/AttributeReaderTest.php
@@ -102,6 +102,17 @@ final class AttributeReaderTest extends TestCase
         );
     }
 
+    public function testDoesNotInstantiateUnrelatedAttributes(): void
+    {
+        $reader = new AttributeReader();
+        $class = new \ReflectionClass(EntityWithUnrelatedAttribute::class);
+
+        $this->assertEquals(
+            [AttributeUploadable::class => new AttributeUploadable()],
+            $reader->getClassAttributes($class)
+        );
+    }
+
     /**
      * Test new Attribute namespace methods work correctly.
      */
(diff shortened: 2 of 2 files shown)


garak:
Please rebase

ousamabenyounes:
Rebased onto current `master` as requested (`50b58a9` → `aac6129`). The rebase was conflict-free and retains the original focused fix.

Validation: the regression test is RED on upstream with the expected `LogicException`, then GREEN at 9 tests / 13 assertions; the full local suite passes 348 tests / 1,278 assertions with 2/2 changed lines covered. PHP-CS-Fixer is clean, PHPStan is iso-baseline locally, and every GitHub check is green across PHP 8.1–8.5.

garak (approved):

PR #1605, opened 2026-08-10
🐛 Do not erase file properties when the entity is being removed

### LINKED ISSUE #1117: Add support for not nullable mappings
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes/no
| BC Break    | no

#### Summary

I am using an entity that acts like a file object. Simple representation here:

```php
class MediaObject
{
  /** @Vich\UploadableField(mapping="media_object", fileNameProperty="filename") */
  private ?SymfonyFile $file;

  protected string $filename;
  // ...
  public function __construct() {
    $this->file = null;
    $this->filename = '';
  }
  // ...
  public function setFilename(string $filename) {
    $this->filename = $filename;
  }
  // ...
}
```

This object is only valid as long as the file exists. So if I now want to remove the entity:
```php
$em->remove($mediaObject);
```

I will be greeted with this Exception: `Expected argument of type "string", "null" given at property path "filename".`.

The question now is:
- Do we update the current logic to handle non-nullable properties like:
```php
if ($property->isNullable()) {
  $this->getAccessor()->setValue($propertyName, null);
}
```
- Do we create a new Annotation or mechanic that disables setting of null.

ousamabenyounes:
### Failing Test

| Q | A |
| --- | --- |
| BC Break | no |
| Version | 2.x (master) |

#### Summary

Fixes #1117.

`PropertyMapping::erase()` resets mapped file properties to `null`. When a mapped target is non-nullable, such as `setFilename(string $filename)` or a public typed property, Symfony's property accessor raises a `TypeError`.

#### Change

Add `PropertyMapping::isNullable()` and use it to erase only mapped targets that can accept `null`. The check follows the property accessor's write path: a public camelized setter first, then a public property. Unconfigured, indexed, dynamic, or otherwise non-introspectable targets retain the previous behavior.

This keeps property erasure generic, including during entity deletion, while avoiding invalid null writes to non-nullable targets.

#### Test verification (RED → GREEN)

On unmodified `master` with the regression tests applied:

```text
4 errors, 1 failure across 18 focused tests
```

With the fix:

```text
18 focused tests passed
```

The full PHP 8.3 local CI replay is iso-baseline: `master` ran 348 tests and this branch ran 353 tests, with identical counts of 1 deprecation, 132 notices, 32 skipped tests, and 15 risky tests; neither run had failures.

PHP-CS-Fixer verification for the review follow-up: 1 fixable file before → 0 fixable files after.

### DIFF
--- a/src/Mapping/PropertyMapping.php
+++ b/src/Mapping/PropertyMapping.php
@@ -5,6 +5,7 @@ namespace Vich\UploaderBundle\Mapping;
 use Symfony\Component\HttpFoundation\File\File;
 use Symfony\Component\PropertyAccess\PropertyAccess;
 use Symfony\Component\PropertyAccess\PropertyAccessor;
+use Symfony\Component\PropertyAccess\PropertyPath;
 use Vich\UploaderBundle\Naming\DirectoryNamerInterface;
 use Vich\UploaderBundle\Naming\NamerInterface;
 use Vich\UploaderBundle\Util\PropertyPathUtils;
@@ -16,6 +17,12 @@ use Vich\UploaderBundle\Util\PropertyPathUtils;
  */
 final class PropertyMapping
 {
+    private const UNKNOWN_PROPERTY_MESSAGE = 'Unknown property %s';
+    private const SETTER_PREFIX = 'set';
+    private const PROPERTY_WORD_SEPARATORS = ['_', '-'];
+    private const PROPERTY_WORD_SEPARATOR = ' ';
+    private const EMPTY_STRING = '';
+
     private ?NamerInterface $namer = null;
 
     private ?DirectoryNamerInterface $directoryNamer = null;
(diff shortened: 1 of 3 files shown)


garak (changes requested):
What if, instead, we add a PropertyMapping method to check if the property is actually nullable, and erase it only when we can?
We could keep the erasing for deletion (it doesn't hurt, even if not really useful), and keep the thing more generic.

ousamabenyounes:
Thanks for the review @garak — agreed, that's cleaner and more generic. Reworked in a7545db.

Instead of threading a flag from the remove listener, `PropertyMapping::erase()` now erases each property only when it can actually hold null. The new `PropertyMapping::isNullable()` mirrors the property accessor write path — a public camelized setter first, then a public typed property — so a non-nullable target (the #1117 case) is skipped while nullable ones are still cleared. Erasing on deletion is kept. The previous `$eraseProperties` flag on `UploadHandler::remove()` and the listener change are reverted.

Added `NotNullableArticle` fixture plus tests covering the skip, the direct `isNullable()` result, snake_case paths and public-property writes.

garak:
Thanks for the cooperation.
Please squash your commits and rebase from master (I fixed the Gaufrette-related problems)

ousamabenyounes:
Rebased and squashed the branch onto the latest master, then force-pushed 3f576de. StyleCI currently reports a failure on the new commit; I am following up on it before considering this complete.

ousamabenyounes:
Fixed the remaining PHP-CS-Fixer/StyleCI issue in `3817e5f` by using the configured anonymous-class parentheses style.

Validation: PHP-CS-Fixer RED (1 fixable file) → GREEN (0 fixable files). Local PHP 8.3 CI replay remained iso-baseline: master 348 tests vs. PR 353 tests, with identical deprecation/notice/skip/risky counts and no failures. Remote checks will run on the pushed SHA.

garak (approved):

PR #1610, opened 2026-09-06
Fix built-in namer configuration isolation for 2.10

endelwar:
# Isolate built-in namer configuration between mappings

Mapping options currently mutate shared namer services. A mapping that omits an option can therefore inherit the value configured by another mapping:

```yaml
vich_uploader:
    db_driver: orm
    mappings:
        product_image:
            upload_destination: '%kernel.project_dir%/public/uploads/products'
            namer:
                service: Vich\UploaderBundle\Naming\OrignameNamer
                options: { transliterate: true }

        user_avatar:
            upload_destination: '%kernel.project_dir%/public/uploads/avatars'
            namer: Vich\UploaderBundle\Naming\OrignameNamer
```

Uploading `Fôo Bàr.txt` through `user_avatar` produces different names depending on which mapping was resolved first (random prefix omitted):

```text
Fôo Bàr.txt    when user_avatar is resolved first
foo-bar.txt   when product_image was resolved first
```

**This affects all built-in file and directory namers.** Their `configure()` methods retain options they do not receive: `transliterate`, hash `algorithm` and `length`, directory depth and date formats can all carry over between mappings. Passing some options does not prevent omitted ones from leaking.

The same applies to `keep_extension`: one mapping opting into client-provided extensions can make another mapping retain them too, despite never enabling that option. Resolution also happens during form rendering, URI generation and file removal, so an upload to the first mapping is not required to trigger the leak.

This change copies built-in namers before applying mapping options. Copies preserve service-level configuration and are also created for mappings without options, so previously resolved mappings remain unaffected.

To preserve compatibility in 2.10.x, cloning applies only to the concrete built-in classes. Custom namers, subclasses and decorators retain their existing behavior; their shared mutable configuration remains outside this fix. No public interfaces or container configuration change.


garak:
Please re-target to the 2.x branch

garak (approved):

PR #1611, opened 2026-09-06
Give each mapping its own namer configuration

endelwar:
# Give each mapping its own namer configuration

## The problem

Mappings referencing the same shared namer service use the same instance. The resolver calls
`configure()` on that instance, so options set by one mapping can carry over to another.

```yaml
vich_uploader:
    mappings:
        product_image:
            upload_destination: '%kernel.project_dir%/public/uploads/products'
            namer:
                service: Vich\UploaderBundle\Naming\OrignameNamer
                options: { transliterate: true }

        user_avatar:
            upload_destination: '%kernel.project_dir%/public/uploads/avatars'
            namer: Vich\UploaderBundle\Naming\OrignameNamer
```

Uploading `Fôo Bàr.txt` to `user_avatar` preserves the original name after the generated prefix
when that mapping is resolved first, but produces `<prefix>_foo-bar.txt` if `product_image`
was resolved earlier in the same process. Resolving mappings also happens during form rendering,
URI resolution and file removal, so an upload is not required to trigger the configuration change.

This affects configurable file and directory namers generally: transliteration, hash algorithms,
name lengths, directory layouts and date formats can all leak between mappings. It also affects
`keep_extension`: a mapping can preserve the client-provided extension because another mapping
enabled that option, even though it did not request this behavior itself.

`ChainDirectoryNamer` shares both the chain instance and its configurable children. For example:

```yaml
vich_uploader:
    mappings:
        invoice:
            upload_destination: '%kernel.project_dir%/public/uploads/invoices'
            directory_namer:
                service: Vich\UploaderBundle\Naming\ChainDirectoryNamer
                options:
                    namers:
                        - service: Vich\UploaderBundle\Naming\CurrentDateTimeDirectoryNamer
                          options: { date_time_property: createdAt, date_time_format: 'Y-m' }

        product_image:
            upload_destination: '%kernel.project_dir%/public/uploads/products'
            directory_namer:
                service: Vich\UploaderBundle\Naming\ChainDirectoryNamer
                options:
                    namers:
                        - service: Vich\UploaderBundle\Naming\CurrentDateTimeDirectoryNamer
                          options: { date_time_property: createdAt }
```

For an entity created on 2024-01-15, `product_image` uses `2024/01/15` when resolved first,
but inherits `2024-01` after `invoice` is resolved. Resolving another chain can also replace
the children or separator used by an already resolved mapping.

## The fix

`ConfigurableInterface` now declares `withOptions(array $options): static` instead of
`configure()`. The resolver uses the returned instance for the mapping and leaves the shared
service unchanged. This applies to built-in and custom configurable namers, including subclasses
and decorators.

- Each call must return a new instance with independent mutable configuration, preserving service
  defaults. Empty options still produce a separate instance with the defaults at resolution time.
- Chains copy their configurable children, including those supplied directly to `withOptions()`.
  The resolver supports nested chains and repeated services with different options.
  Non-configurable children remain shared. Explicit `namers: []` clears the chain.
- Returning `$this` raises a `LogicException`. Implementations remain responsible for isolating
  any mutable configuration held in dependencies or nested objects.

The eleven other built-in namers use `ConfigurableNamerTrait`, which clones the instance and
calls `configure()` on the copy when options are non-empty. Their `configure()` methods remain
available for service-level defaults, but are no longer required by the interface.

## Upgrading a custom namer

Existing mapping configuration needs no changes. Custom implementations of `ConfigurableInterface`
must provide `withOptions()`. When a shallow clone isolates the configuration, adding the trait
to the existing class is sufficient:

```php
class MyNamer implements NamerInterface, ConfigurableInterface
{
    use \Vich\UploaderBundle\Naming\ConfigurableNamerTrait;

    // Existing configure() and name() implementations.
}
```

For mutable configuration objects, copy them in `__clone()` or implement `withOptions()` by
constructing a new instance. Decorators must obtain an independent inner namer and wrap it,
for example `return new self($this->inner->withOptions($options));`. A `readonly` property alone
does not make the object it holds immutable.

Migration details are included in `UPGRADE.md` and the custom namer guides.

## Relation to the 2.10 patch

The proposed 2.10.x patch isolates the eleven built-in namer classes without changing the public
interface. This 3.x change introduces an explicit contract for custom configurable namers,
subclasses and decorators, and handles chains and their configurable children recursively.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,10 @@
+# Upgrading from v3.0 to v3.1
+
+## Deprecations
+
+* `ConfigurableInterface` is deprecated in favor of `ImmutableConfigurableInterface`, which declares `withOptions(array $options): static` instead of `configure(array $options): void`. Namer services are shared, so configuring one in place makes every mapping observe the options of the last one resolved. `withOptions()` must return a new instance with independent mutable configuration, preserving service defaults; it is called for every configurable namer, including those inside chains and mappings without options. `ConfigurableNamerTrait` implements it when a shallow clone is enough; see the [custom namer guide](docs/file_namer/howto/create_a_custom_file_namer.md#configurable-custom-namer).
+* Namers implementing only `ConfigurableInterface` keep working, without configuration isolation, until its removal in 4.0. A `configure()` method is still usable for service-level defaults.
+
 # Upgrading from v2.9 to v3.0
 
 ## Breaking Changes
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,7 @@
         "symfony/config": "^6.4 || ^7.4 || ^8.0",
         "symfony/console": "^6.4 || ^7.4 || ^8.0",
         "symfony/dependency-injection": "^6.4 || ^7.4 || ^8.0",
+        "symfony/deprecation-contracts": "^3.0",
         "symfony/event-dispatcher-contracts": "^3.1",
         "symfony/http-foundation": "^6.4 || ^7.4 || ^8.0",
         "symfony/http-kernel": "^6.4 || ^7.4 || ^8.0",
--- a/docs/directory_namer/howto/create_a_custom_directory_namer.md
+++ b/docs/directory_namer/howto/create_a_custom_directory_namer.md
@@ -20,6 +20,12 @@ vich_uploader:
             directory_namer: my.directory_namer.product
 ```
 
+## Configurable directory namers
+
+To accept mapping options, also implement `ImmutableConfigurableInterface`. Directory namers use the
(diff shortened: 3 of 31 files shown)


garak:
Please re-target to master

endelwar:
Heads-up: now that v3.0.0 is tagged, this PR can't land as-is, it changes `ConfigurableInterface::configure(array): void` into `withOptions(array): static`, which breaks every third-party namer implementing that interface, so it's no longer eligible for a 3.0.x/3.1.x release.

I'm reworking it into a BC-safe variant: a new `ImmutableConfigurableInterface` carrying `withOptions()`, `ConfigurableInterface` left untouched and deprecated, with the resolver falling back to the legacy `configure()` path. I'll push the update here shortly.

garak (approved):

